PR #40, opened 2019-04-17
Specify channel in conda installation instructions

datapythonista:
Making explicit that when installing from conda, static-frame needs to use conda-forge channel. I guess it's assumed to be in the user settings, but that's probably not the case for most users.


flexatone:
Thanks for this helpful change, I will merge it in shortly.

PR #47, opened 2019-05-05
Exclude test and performance folders form coverage

gliptak:


flexatone:
Thank you for this refinement!

PR #51, opened 2019-05-14
Build and test against Windows and macOS.

brandtbucher:
@flexatone: It looks like a lot of our tests are sensitive to platform-specific `dtype` choices. For example, `numpy` uses a C `long` for integer types. This is an `int64` on macOS and Linux, but an `int32` on Windows.

Uses the template from https://github.com/brandtbucher/travis-python-matrix.


brandtbucher:
I've fixed several trivial failures by explicitly setting the `dtype=np.int64`. Everything else (except the doctest failure, since that one is higher-priority) has been conditionally skipped on Windows for now.

PR #61, opened 2019-06-11
python -m static_frame

brandtbucher:
Now, running `python -m static_frame` or `ipython -m static_frame` drops you into an interactive interpreter with `sf`, `np`, and (if installed) `pd` pre-loaded. Here is what that looks like (in each case, the user hasn't interacted yet):

```
$ python -m static_frame
Python 3.7.3 (default, Jun  7 2019, 15:55:34) 
[GCC 5.4.0 20160609] on linux
>>> import numpy as np  # 1.15.1
>>> import pandas as pd  # 0.23.4
>>> import static_frame as sf  # 0.3.5-dev
>>> 
```

`IPython` gets to keep its colored output (can't see it here, obviously):

```
$ ipython -m static_frame

------> import numpy as np  # 1.15.1
------> import pandas as pd  # 0.23.4
------> import static_frame as sf  # 0.3.5-dev

Python 3.7.3 (default, Jun  7 2019, 15:55:34) 
Type 'copyright', 'credits' or 'license' for more information
IPython 7.4.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: 
```

### DIFF
--- a/static_frame/__main__.py
+++ b/static_frame/__main__.py
@@ -1,0 +1,69 @@
+'''Drop into an interactive interpreter pre-loaded with sf, np, and (if installed) pd.
+
+$ python -m static_frame
+$ ipython -m static_frame
+'''
+
+
+from code import interact
+from sys import platform
+from sys import version
+
+import numpy as np
+
+import static_frame as sf
+
+
+imports = {'np': np, 'sf': sf}
+
+
+try:  # Import pandas, if it's installed:
+    import pandas as pd
+except ImportError:
+    pass
+else:
+    imports['pd'] = pd
+
+
+commands = sorted(
+        f'import {package.__name__} as {name}  # {package.__version__}'
+        for name, package in imports.items()
+)
+
+
+try:  # This lets us play nicely with IPython:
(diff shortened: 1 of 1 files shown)


brandtbucher:
It looks like `asyncio` does something like this in 3.8: https://github.com/python/cpython/pull/13472.

In the future, it could also be useful to allow CSVs (or other files) to be passed in as command-line args and pre-loaded as well:

```
$ python -m static_frame somewhere/some_data.csv somewhere/some_other_data.tsv
Python 3.7.3 (default, Jun  7 2019, 15:55:34) 
[GCC 5.4.0 20160609] on linux
>>> import numpy as np  # 1.15.1
>>> import pandas as pd  # 0.23.4
>>> import static_frame as sf  # 0.3.5-dev
>>> some_data = sf.Frame.from_csv("somewhere/some_data.csv")
>>> some_other_data = sf.Frame.from_csv("somewhere/some_other_data.csv")
>>>
```

flexatone:
Thanks for this nice addition!

Issue #62, opened 2019-06-11
Implement __reversed__ in Index, Series, and Frame


flexatone:

colinalexander:
Completed per PRs https://github.com/InvestmentSystems/static-frame/pull/68 & https://github.com/InvestmentSystems/static-frame/pull/69

PR #66, opened 2019-06-16
Index.__reversed__

colinalexander:
In response to Issue https://github.com/InvestmentSystems/static-frame/issues/62, the following implements `__reversed__` for an Index and provides a unit test.  Assuming this PR is acceptable, I will complete the same for both a Series and a Frame.

Given that an Index has methods for both `__len__` and `__getitem__`, executing `reversed(Index)` will already work.  This PR, however, results in significantly faster execution of that function.

### DIFF
--- a/static_frame/core/index.py
+++ b/static_frame/core/index.py
@@ -286,18 +286,6 @@ class Index(IndexBase,
             # resolving the detype is expensive
             labels = collection_to_array(labels, dtype=dtype, discover_dtype=True)
 
-            # if not len(labels):
-            #     return EMPTY_ARRAY # already immutable
-
-            # if isinstance(labels[0], tuple):
-            #     # special handling for iterables of
-            #     assert dtype is None or dtype == object
-            #     array = np.empty(len(labels), object)
-            #     array[:] = labels
-            #     labels = array # rename
-            # else:
-            #     labels = np.array(labels, dtype)
-
         else: # labels may be an expired generator, must use the mapping
 
             # TODO: explore why this does not work
@@ -750,12 +738,6 @@ class Index(IndexBase,
         '''
         return self._map.keys()
 
-    def __iter__(self):
-        '''Iterate over labels.
-        '''
-        if self._recache:
-            self._update_array_cache()
-        return self._labels.__iter__()
 
     def __contains__(self, value) -> bool:
         '''Return True if value in the labels.
--- a/static_frame/core/index.py
+++ b/static_frame/core/index.py
@@ -430,6 +430,12 @@ class Index(IndexBase,
         # NOTE: automatic discovery is possible, but not yet implemented
         self._loc_is_iloc = loc_is_iloc
 
+    def __reversed__(self) -> tp.Iterator[tp.Hashable]:
+        '''
+        Returns a reverse iterator on the index labels.
+        '''
+        return reversed(self.values)
+
     #---------------------------------------------------------------------------
     def __setstate__(self, state):
         '''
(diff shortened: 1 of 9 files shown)


static_frame/core/index.py
```diff
@@ -430,6 +430,13 @@ def __init__(self,
         # NOTE: automatic discovery is possible, but not yet implemented
         self._loc_is_iloc = loc_is_iloc
 
+    def __reversed__(self):
```
brandtbucher: I'm gradually trying to get the codebase to pass static type checking, so a type hint here would be great:

```suggestion
    def __reversed__(self) -> tp.Iterator[tp.Any]:
```

(`Index` isn't `Generic` yet, so `tp.Any` as the item type is fine for now.)

static_frame/core/index.py
```diff
@@ -430,6 +430,13 @@ def __init__(self,
...
+        '''
+        Returns a reverse iterator on the index labels.
+        '''
+        for label in self.values[::-1]:
```
brandtbucher: I ran some time trials, and it looks like simply returning a reverse iterator of the underlying array is more than twice as fast to iterate over (it's a bit slower to call initially, but that difference is insignificant when compared to iteration time). So this should be fine:

```suggestion
        return reversed(self.values)
```

static_frame/core/index.py
```diff
@@ -430,6 +430,13 @@ def __init__(self,
...
+            yield label
```
brandtbucher: ...and then this can be removed:
```suggestion
```

static_frame/core/index.py
colinalexander: Done.

static_frame/core/index.py
colinalexander: Done.

colinalexander:
1 build error which appears unrelated to the PR:
```
    distutils.errors.DistutilsError: Download error for https://files.pythonhosted.org/packages/e0/27/a913b227c9053472ea7dda33e409a1e748e50cdd208a955f3d6a20e9a875/pbr-5.3.0-py2.py3-none-any.whl#sha256=089ccb087e9bd8f278caedfa6c2c5d461381437eda3db750b6834e78b319f404: [Errno 8] nodename nor servname provided, or not known
    
    ----------------------------------------
Command "python setup.py egg_info" failed with error code 1 in /private/var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/pip-install-awn9yqcf/sphinx-jinja/
The command "pip install -r requirements-test.txt" failed and exited with 1 during .```

**Comment by brandtbucher** [comment]
Thanks for the heads-up. I’ve restarted the job.

File: static_frame/core/index.py [hunk 4]
```diff
@@ -430,6 +430,12 @@ def __init__(self,
         # NOTE: automatic discovery is possible, but not yet implemented
         self._loc_is_iloc = loc_is_iloc
 
+    def __reversed__(self) -> tp.Iterator[tp.Hashable]:
+        '''
+        Returns a reverse iterator on the index labels.
+        '''
+        return reversed(self.values)
```
brandtbucher: On second thought, it's probably better to bypass the descriptor altogether and just do what `__iter__` is doing:

```suggestion
        if self._recache:
            self._update_array_cache()
        return reversed(self._labels)
```

flexatone:
Hi @colinalexander, many thanks for this PR!

I think we can implement this in a manner that works for both `Index` and `IndexHierarchy`. In thinking about this, I found the opportunity to move the implementation of `__iter__` to `IndexBase`, instead of keeping identical implementations in both `Index` and `IndexHierarchy`. I have made this change.

Can you try implementing `__reversed__` on `IndexBase`, following the approach of `IndexBase.__iter__`, as @brandtbucher suggested? Using `.values` is slightly heavier then using `_labels` directly, so I would favor the latter approach.

PR #67, opened 2019-06-17
Removing one instance of package listed twice

colinalexander:
`sphinx-jinja==0.3.0` is listed twice in the `requirements-test.txt` file.  Possibly related to the build failure earlier, but certainly worth cleaning up in any case.


flexatone:
Thanks for this refinement!

PR #68, opened 2019-06-17
Series.__reversed__ / Frame.__reversed__

colinalexander:
A continuation of PR https://github.com/InvestmentSystems/static-frame/pull/66 (responding to Issue https://github.com/InvestmentSystems/static-frame/issues/62), the following implements` __reversed__` for both a Series and a Frame and provides unit tests.

### DIFF
--- a/static_frame/core/frame.py
+++ b/static_frame/core/frame.py
@@ -1975,6 +1975,12 @@ class Frame(metaclass=MetaOperatorDelegate):
     #---------------------------------------------------------------------------
     # transformations resulting in the same dimensionality
 
+    def __reversed__(self) -> tp.Iterator[tp.Hashable]:
+        '''
+        Returns a reverse iterator on the frame's columns.
+        '''
+        return reversed(self._columns)
+
     def sort_index(self,
             ascending: bool = True,
             kind: str = DEFAULT_SORT_KIND) -> 'Frame':
--- a/static_frame/core/series.py
+++ b/static_frame/core/series.py
@@ -265,6 +265,13 @@ class Series(metaclass=MetaOperatorDelegate):
         if len(self.values) != shape:
             raise RuntimeError('values and index do not match length')
 
+    # ---------------------------------------------------------------------------
+    def __reversed__(self) -> tp.Iterator[tp.Hashable]:
+        '''
+        Returns a reverse iterator on the series' index.
+        '''
+        return reversed(self._index)
+
     #---------------------------------------------------------------------------
     def __setstate__(self, state):
         '''
--- a/static_frame/test/unit/test_frame.py
+++ b/static_frame/test/unit/test_frame.py
@@ -1830,6 +1830,20 @@ class TestUnit(TestCase):
                 ((('i', 'a'), (('a', 'a'), ('b', 'b'), ('c', 'c'))), (('i', 'b'), (('a', 'tru'), ('b', 'fals'), ('c', 'tru'))), (('ii', 'a'), (('a', 'non'), ('b', 'non'), ('c', 'non'))), (('ii', 'b'), (('a', 'non'), ('b', '1'), ('c', '5'))))
                 )
 
+    def test_frame_reversed(self):
+        columns = tuple('pqrst')
+        index = tuple('zxwy')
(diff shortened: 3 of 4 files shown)


static_frame/core/frame.py
```diff
@@ -1975,6 +1975,12 @@ def _iter_element_loc(self):
     #---------------------------------------------------------------------------
     # transformations resulting in the same dimensionality
 
+    def __reversed__(self) -> tp.Iterable[np.ndarray]:
```
brandtbucher:
```suggestion
    def __reversed__(self) -> tp.Iterator[tp.Hashable]:
```

static_frame/core/frame.py
```diff
@@ -1975,6 +1975,12 @@ def _iter_element_loc(self):
...
+        '''
+        Returns a reverse iterator on the frame's rows.
+        '''
+        return reversed(self.values)
```
brandtbucher: In both of these cases, I think we want to iterate over the index, like a dictionary. Basically, a reversed version of `iter(frame)` or `iter(series)`. So this would give the desired behavior:

```suggestion
        return reversed(self.index)
```

static_frame/core/series.py
```diff
@@ -265,6 +265,13 @@ def values_constructor(shape):
         if len(self.values) != shape:
             raise RuntimeError('values and index do not match length')
 
+    # ---------------------------------------------------------------------------
+    def __reversed__(self) -> tp.Iterable[np.ndarray]:
```
brandtbucher: Here too:

```suggestion
    def __reversed__(self) -> tp.Iterator[tp.Hashable]:
```

static_frame/core/series.py
```diff
@@ -265,6 +265,13 @@ def values_constructor(shape):
...
+        '''
+        Returns a reverse iterator on the series' values.
+        '''
+        return reversed(self.values)
```
brandtbucher: And also here:

```suggestion
        return reversed(self.index)
```

colinalexander:
Not that the Pandas implementation is correct, but I just wanted to point out the inconsistency between calling `reversed` on a pandas series vs a static_frame series per the requested modification `reversed(self._index)`:

```
sf_series = sf.Series(range(4), index=('a', 'b', 'c', 'd'))
>>> tuple(reversed(sf_series))
('d', 'c', 'b', 'a')

pd_series = pd.Series(range(4), index=('a', 'b', 'c', 'd'))
>>> tuple(reversed(pd_series))
(3, 2, 1, 0)
```

I'm happy to make the change but just wanted you to be aware of the difference.

FYI, `reversed(df)` leads to `KeyError` in pandas.

```
# Pandas reversed dataframe.
df = pd.DataFrame(np.random.randn(4, 4), columns=list('abcd'))
>>> reversed(df).__next__()
KeyError: 3

# Proposed static-frame implementation.
f = sf.Frame(np.random.randn(4, 4), columns=list('abcd'))
>>> tuple(reversed(f))
('d', 'c', 'b', 'a')
```

brandtbucher:
I think that SF's approach is better. The [docs for `__reversed__`](https://docs.python.org/3/reference/datamodel.html#object.__reversed__) say that it should implement "reverse iteration", which I take to mean "the reverse of `__iter__`".

In fact, Python 3.8 gives dictionaries `reversed` support... and it iterates over the keys, too!

```py
>>> test = {'a': 1, 'b': 2, 'c': 3}
>>> tuple(test)
('a', 'b', 'c')
>>> tuple(iter(test))
('a', 'b', 'c')
>>> tuple(reversed(test))
('c', 'b', 'a')
```

colinalexander:
Yes, that's fine.  I've made the changes per your approach.

flexatone:
Thanks, @colinalexander, for implementing this PR.

As @brandtbucher states, we have taken the position of deviating with Pandas as to how a `Series` iterates to be consistent with `Frame` and the overall dict-like interface of these containers. This has the side-effect of forcing users to use `.values` when when they want to expose an iterable of values, but we find that this too is positive, as it makes explicit what is happening, and delegates to `__iter__` on the underlying (immutable) NumPy array.

PR #112, opened 2019-10-01
Separate requirements.txt file

gliptak:
Signed-off-by: Gábor Lipták <gliptak@gmail.com>


flexatone:
Many thanks for this PR. I had been meaning to do something like this for some time!

Issue #9, opened 2018-10-26
Series.isin(): performance


flexatone:

flexatone:
`Index.isin()` seems particularly slow in some cases too.

brandtbucher:
I'm gonna reopen this following our discussion. @chaburkland

PR #128, opened 2019-11-13
isin performance refactor

chaburkland:
Resolves Issue #9 

1) Static Frame's current behavior for `isin` calls is to use numpy's built-in `isin` method. This presents a performance issue, as numpy is not optimized to work with object arrays. It results in a time complexity of O(n^2).
2) This PR resolves this by adding a branch within the `isin` logic which converts the lookup argument to a set instead of a array-like structure. This is to take advantage of a set's constant lookup time, reducing the time complexity to O(n).
3) This optimization only applies when either the callee's underlying data or the lookup argument contain objects, as numpy is already optimized for primitive type arguments.
4) If either the callee's underlying data or the lookup argument contain unhashable data (i.e. conversion to a set will fail), it will fall back on numpy's `isin` functionality.
5) Adds unit and property tests for new public methods.
6) Adds benchmark comparisons.

### DIFF
--- a/static_frame/core/frame.py
+++ b/static_frame/core/frame.py
@@ -2,7 +2,7 @@ import typing as tp
 import sqlite3
 import csv
 import json
-from collections import namedtuple
+from collections import namedtuple, abc
 from functools import partial
 from itertools import chain
 from itertools import repeat
@@ -44,6 +44,7 @@ from static_frame.core.util import column_1d_filter
 from static_frame.core.util import name_filter
 from static_frame.core.util import _gen_skip_middle
 from static_frame.core.util import iterable_to_array
+from static_frame.core.util import isin
 # from static_frame.core.util import _dict_to_sorted_items
 from static_frame.core.util import array_to_duplicated
 from static_frame.core.util import ufunc_set_iter
@@ -3371,11 +3372,7 @@ class Frame(ContainerOperand):
         '''
         Return a same-sized Boolean Frame that shows if the same-positioned element is in the iterable passed to the function.
         '''
-        # cannot use assume_unique because do not know if values are unique
-        v, _ = iterable_to_array(other)
-        # NOTE: is it faster to do this at the block level and return blocks?
-        array = np.isin(self.values, v)
-        array.flags.writeable = False
+        array = isin(self.values, other)
         return self.__class__(array, columns=self._columns, index=self._index)
 
     @doc_inject(class_name='Frame')
@@ -3383,7 +3380,7 @@ class Frame(ContainerOperand):
             lower=None,
             upper=None,
             axis: tp.Optional[int] = None):
-        '''{}
+        '''
 
         Args:
             lower: value, ``Series``, ``Frame``
--- a/static_frame/core/frame.py
+++ b/static_frame/core/frame.py
@@ -2,7 +2,7 @@ import typing as tp
 import sqlite3
 import csv
 import json
-from collections import namedtuple, abc
+from collections import namedtuple
 from functools import partial
 from itertools import chain
 from itertools import repeat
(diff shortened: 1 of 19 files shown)


static_frame/core/index_hierarchy.py
```diff
@@ -24,6 +24,8 @@
 from static_frame.core.util import array2d_to_tuples
 from static_frame.core.util import name_filter
 from static_frame.core.util import iterable_to_array
```
brandtbucher: This import is unused now:

```suggestion
```

static_frame/core/frame.py
```diff
@@ -3371,19 +3372,15 @@ def isin(self, other) -> 'Frame':
...
+        array = isin(self.values, other)
         return self.__class__(array, columns=self._columns, index=self._index)
 
     @doc_inject(class_name='Frame')
     def clip(self,
             lower=None,
             upper=None,
             axis: tp.Optional[int] = None):
-        '''{}
+        '''
```
brandtbucher: These brackets are significant, they're filled by the `@doc_inject` decorator above.

```suggestion
        '''{}
```

static_frame/core/index_hierarchy.py
```diff
@@ -943,24 +946,14 @@ def isin(self, other: tp.Iterable[tp.Iterable[tp.Hashable]]) -> np.ndarray:
         for seq in other:
             if not hasattr(seq, '__iter__'):
                 raise RuntimeError('must provide one or more iterables within an iterable')
-            # must use iterable to array to properly handle heterogenous types, or if already an array
-            v, _ = iterable_to_array(seq)
-            # if seq is a tuple, could check self.keys()
-            if len(v) == self.depth:
-                matches.append(v)
+            if len(seq) == self.depth:
```
brandtbucher: I think we still need lines 946-949, since `seq` could be an unsized iterator.

static_frame/core/util.py
```diff
@@ -1570,6 +1570,101 @@ def ufunc_set_iter(
...
+    other, assume_unique = iterable_to_array(other)
+
+    if array.dtype == DTYPE_OBJECT or other.dtype == DTYPE_OBJECT:
+        try:
+            if array.ndim == 1:
+                result = _isin_1d(array, set(other))
+            else:
+                result = _isin_2d(array, set(other))
+        except TypeError:
+            # TypeErrors *should* only occur when something is unhashable, hence the inability to use sets. Fall back to numpy's isin.
```
brandtbucher: Not sure I like this. Perhaps just convert to `set` in the `try` block, then call the helper functions in the `else`?

chaburkland: The reason for this is that there might be unhashable elements in `array`, not just `other`. This will raise a TypeError when `= element in other` is called, since `element` is coming from `array`.

static_frame/core/index_hierarchy.py
chaburkland: That is a good point. I will fix that.

static_frame/core/util.py
brandtbucher: Hm, okay.

brandtbucher: It may be worth having the helpers perform the `set` conversion themselves and return `None` or something if there are any issues with that, just to avoid swallowing all `TypeError`s like this.

But it may be fine as-is. Not sure.

chaburkland: I considered that too, but it introduces even more overhead by adding another iteration over `array`'s contents. I didn't think it was worth the cost as my assumption is that unhashable elements are going to be a rare use case.

Ultimately, it seems like a judgment call of, do we accept overhead now, or later? Since unhashable elements are likely a significantly minor percentage of use cases, it seems appropriate to dump the overhead on them, instead of the normal case.

flexatone: Agreed, unhashable elements are likely a minority of cases.

Issue #135, opened 2019-11-25
Series/Frame.from_concat() raises on empty iterable


chaburkland:
`Frame.from_concat()` and `Series.from_concat()` both raise `StopIteration` exceptions when passed an empty iterable. One of two things should happen instead of this:

1) Return an empty Series/Frame when `from_concat()` is called with an empty iterable.
2) Raise a different exception (or even an assert failure), as `StopIteration` is not clear.

My recommendation is 1), as that feels the most intuitive. If my argument contains no values, I expect to receive a Series/Frame that also contains no values.

brandtbucher:
I'd even go so far as to say that (1) is the *expected* behavior, and this is a bug.

brandtbucher:
This also affects `Frame.from_concat_items` / `Series.from_concat_items`. Fixing it should be as simple as wrapping the call to `next` in a `try`/`except` block...

flexatone:
Thanks to @chaburkland for the fix!

PR #137, opened 2019-11-27
Allow for empty iterable to be used in from_concat and from_concat_items

chaburkland:
Addresses issue #135 
Adds ability to use Series/Frame's `from_concat` and `from_concat_items` with empty iterables.
Adds unit tests to demonstrate the changed behavior.


flexatone:

static_frame/core/series.py
```diff
@@ -187,6 +187,10 @@ def from_concat(cls,
             if index is None:
                 array_index.append(c.index.values)
 
+        # End quickly if empty iterable
+        if len(array_values) == 0:
```
flexatone: if not len(array_values):, perhaps?

static_frame/core/series.py
```diff
@@ -218,10 +222,18 @@ def gen():
...
-        return cls(values, index=ih, own_index=True)
+        try:
+            ih = IndexHierarchy.from_index_items(gen())
+            # returns immutable array
+            values = concat_resolved(array_values)
+            own_index = True
+        except StopIteration:
+            # Default to empty when given an empty iterable
+            ih = None
+            values = []
```
flexatone: Can values be something other than a list, such as EMPTY_TUPLE?

static_frame/test/unit/test_frame.py
```diff
@@ -4618,6 +4618,30 @@ def test_frame_from_concat_w(self) -> None:
                 )
 
 
+    def test_frame_from_concat_x(self) -> None:
+        f1 = Frame.from_concat([])
+        f2 = Frame.from_concat([], columns='a')
+        f3 = Frame.from_concat([], index=[])
+        f4 = Frame.from_concat([], name='f4')
+        f5 = Frame.from_concat([], columns='a', index=[], name='f5')
```
flexatone: The diversity of test inputs here is good, but I think this is more maintainable if the Frames are next to the code that tests them, rather then split into two groups.

static_frame/core/frame.py
```diff
@@ -187,13 +187,23 @@ def from_concat(cls,
         # axis_series = (0 if axis is 1 else 1)
         frames = [f if isinstance(f, Frame) else f.to_frame(axis) for f in frames]
 
+        own_columns = False
+        own_index = False
+
+        # End quickly if empty iterable
+        if len(frames) == 0:
```
flexatone: if not len(frames):, perhaps?

static_frame/core/frame.py
```diff
@@ -187,13 +187,23 @@ def from_concat(cls,
...
+            return cls(
+                    index=index,
+                    columns=columns,
+                    name=name,
+                    own_data=True,
```
flexatone: own_data should only be true if the passed data object is an ownable TypeBlocks instance. Here, it may not matter, but it is best left unset.

static_frame/core/series.py
```diff
@@ -187,6 +187,10 @@ def from_concat(cls,
             if index is None:
                 array_index.append(c.index.values)
 
+        # End quickly if empty iterable
+        if len(array_values) == 0:
+            return cls([], index=index, name=name)
```
flexatone: As below, can we initialize with something other than a list, such as EMPTY_TUPLE?

static_frame/core/frame.py
chaburkland: How about `if not frames:` as frames is a Python list, and list are truthy.

static_frame/test/unit/test_frame.py
chaburkland: Good call. I will rearrange this code.

static_frame/core/frame.py
chaburkland: Thanks. I was a bit unsure of what the parameters meant, so I attempted to mimic the original source as closely as possible. It makes more sense to remove it though.

static_frame/core/frame.py
flexatone: Nice!

Issue #141, opened 2019-12-10
Frame Group Iterators run slowly on large data sets


chaburkland:
I am working with a very large frame (`frame.shape = (20329978, 10)`), and I have found that iterating using any of the four `iter_group` type iterators is very slow.

I believe the performance is bottlenecked by the speed of the loc call. From testing, I've found the speed of `iter_group_items` is nearly equal to:
```python
for group in frame[col_key].unique():
    yield group, frame.loc[frame[col_key]==group][frame.columns!=col_key]
```

This is not an issue when frames are smaller, but as the frame grows, the loc time grows, thus making these calls very slow.

My suggestion is to make a private, internal sorted copy before grouping, and then yield direct iloc slices from the underlying block structure (and index) directly. My brief testing has shown this to radically increase the speed when working with large frames. See the following proof-of-concept sketch:

```python
def groupby(self, key):
    # Sort a copy since we do not want to change our underlying data
    copy = self.sort_values(key)

    MAX_ILOC = len(self)
    COL_ILOC = self._columns.loc_to_iloc(key)
    COL_KEY = self.columns != key
    COLUMNS = self.columns[COL_KEY]

    # Determine starting group
    group = copy.iloc[0, COL_ILOC]

    start = 0
    i = 0

    while i < MAX_ILOC:
        # Yield a new group/frame pair when we encounter a different group
        if group != copy.iloc[i, COL_ILOC]:
            # Build frame
            _slice = slice(start, i)
            index = copy.index[_slice]
            result = copy._blocks._extract_array(row_key=_slice, column_key=COL_KEY)
            yield group, Frame(result, columns=COLUMNS, index=index)

            # Reset
            start = i
            group = copy.iloc[i, COL_ILOC]
        i += 1

    # Return the last group.
    index = copy.index[start:]
    result = copy._blocks._extract_array(row_key=slice(start, None), column_key=COL_KEY)
    yield group, Frame(result, columns=COLUMNS, index=index)
```

The only caveat is that this approach is not optimal for smaller frame sizes. It's feasible that there exists a good threshold indicator for when to change the `iter_group` behavior.

flexatone:
Excellent suggestion. Do you have any idea at what size changing the `iter_group` behavior is optimal?

chaburkland:
[...]

PR #145, opened 2019-12-19
Frame `iter_group` optimization. Frame sort and `iter_group` when axis=1 fix;

### LINKED ISSUE #143: Error using sort_values() when axis=0 and underlying block structure is homogenous.
When working with a frame that has a homogeneous underlying block structure (i.e. a single 2D array), it will raise a TypeError when sorting by index instead of column. However, if the block structure is made up of multiple components, it works as expected.

```python
data = np.array([np.random.randint(i+1) for i in range(9)]).reshape(3,3)
bad_frame = sf.Frame(data, columns=tuple('abc'), index=tuple('xyz'))

assert len(bad_frame._blocks._blocks) == 1
try:
    bad_frame.sort_values('x', axis=0) # Will raise.
except TypeError:
    assert True
else:
    assert False

records = ((4, 2, 3), (2, 3.1, False), (6, False, 3.4))
good_frame = sf.Frame.from_records(records, columns=tuple('abc'), index=tuple('xyz'))

assert len(good_frame._blocks._blocks) > 1
good_frame.sort_values('x', axis=0) # Will run successfully.
assert True
```

chaburkland:
1) Adds optimization for Frame `iter_group` iterators. Addresses #141 
2) Fixes bug where iterating over Frame groups on axis=1 would yield improper results
3) Adds unit tests to verify fix for 2)
4) Fixes bug where sorting on a Frame (when axis=1 and the underlying block structure is homogeneous) will raise an error. Fixes #143 
5) Adds a unit test to verify fix for 4)
6) Adds custom performance comparison module to test Frame `iter_group` iterators for large test frames.
7) Adds BuildTestFrames test class to create different types of large frames to run performance tests on.
8) Fixes a few typos


flexatone:
Thanks for this valuable PR!

PR #180, opened 2020-03-01
Fix Frame and Series from_pandas() pandas >1.0.0 issues

adamczykm:
It might hit performance, but it seems the safest way to support pandas 1.0.0 ExtensionArray.

### DIFF
--- a/static_frame/core/util.py
+++ b/static_frame/core/util.py
@@ -240,9 +240,12 @@ def mloc(array: np.ndarray) -> int:
     return tp.cast(int, array.__array_interface__['data'][0])
 
 
-def immutable_filter(src_array: np.ndarray) -> np.ndarray:
+def immutable_filter(src_array: tp.Union[np.ndarray, 'pandas.PandasArray']) -> np.ndarray:
     '''Pass an immutable array; otherwise, return an immutable copy of the provided array.
     '''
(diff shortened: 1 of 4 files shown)


static_frame/core/util.py
```diff
@@ -240,9 +240,12 @@ def mloc(array: np.ndarray) -> int:
     return tp.cast(int, array.__array_interface__['data'][0])
 
 
-def immutable_filter(src_array: np.ndarray) -> np.ndarray:
+def immutable_filter(src_array: tp.Union[np.ndarray, 'pandas.PandasArray']) -> np.ndarray:
     '''Pass an immutable array; otherwise, return an immutable copy of the provided array.
     '''
+    if hasattr(src_array, 'to_numpy'):  # is pandas.core.arrays.base.ExtensionArray
+        src_array = src_array.to_numpy(copy=True)
+        src_array.flags.writeable = False
```
flexatone: Can we isolate the usage of `to_numpy` to the various `from_pandas()` methods? The `immutable_filter()` function is very low level and called all over the place in situations where we know we already have a NumPy array. It seems preferable to keep this method simple and normalize Pandas data to NumPy arrays upstream.

flexatone:
Not sure if you are able to refine this PR further right now, so I will go ahead and merge it in and then look to some alternative approaches. Thanks again for this contribution.

Issue #118, opened 2019-10-22
IndexBase.difference


flexatone:

flexatone:
Completed with PR #185

PR #185, opened 2020-03-11
Issue #118: Adds index difference method

chaburkland:
Adds logic to handle index differences. (Implements feature request #118)

The only change to the existing logic (outside of my additions), is that when `func` is called now, it is passed the `assume_unique` kwarg for `np.setdiff1d` and `np.intersect1d` calls (`np.union1d` does not have that option). Previously that kwarg was not being passed in. Here is the code snippet that performs that:
```python3
func_kwargs = {} if is_union else dict(assume_unique=assume_unique) 
return func(array, other, **func_kwargs)
```

Other observations:

- Order is not being preserved as the docstrings suggest. The calls to `frozenset` are unordered, which means when that branch is hit, order promises are lost. I refrained from addressing that in this PR in order to keep the changes atomic.
- `nan`s do not seem to be handled correctly. See the following code snippet:
```python3
>>> a = np.array([1, 2, np.nan, np.nan])
>>> b = np.array([3, np.nan, 4, np.nan])
>>> sf.core.util.intersect1d(a, b)
array([], dtype=float64)
>>> sf.core.util.union1d(a, b)
array([ 1.,  2.,  3.,  4., nan, nan, nan, nan])
>>> sf.core.util.setdiff1d(a, b)
array([ 1.,  2., nan, nan])
```

### DIFF
--- a/static_frame/core/index.py
+++ b/static_frame/core/index.py
@@ -42,6 +42,7 @@ from static_frame.core.selector_node import InterfaceGetItem
 from static_frame.core.selector_node import InterfaceSelection1D
 from static_frame.core.util import union1d
 from static_frame.core.util import intersect1d
+from static_frame.core.util import setdiff1d
 from static_frame.core.util import to_datetime64
 from static_frame.core.util import INT_TYPES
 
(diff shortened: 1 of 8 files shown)


flexatone:
Many thanks for this excellent PR. The implementation is great, and I very much appreciate the extensive unit and property tests.

PR #224, opened 2020-07-01
Fixes type_blocks reference cycle.

chaburkland:
There is a sneaky reference cycle happening in the creation of type_blocks. Essentially, every type_blocks instance has an attribute `iloc` that contained a reference to an instance of `InterfaceGetItem`, which in turn has a reference to the type_blocks instance itself. This meant that when a frame was deleted, it's underlying type_block structure would hang around, since it was still referenced by the `InterfaceGetItem` instance, which itself was referenced by the type_blocks instance, which was reference by the `InterfaceGetItem` instance, and on and on and on. (Many thanks to @brandtbucher for helping to isolate this bug).

Here is an interesting example of how this reference cycle could lead to memory issues:
```python3
import gc
size = 10_000

f = sf.Frame(np.arange(size*size).reshape(size, size))

sel = np.full(size, True)

for _ in range(20): # Each additional iteration adds ~1GB of additional memory.
    f = f.loc[sel]

gc.collect() # This will free ~20GBs of memory, by isolating and deleting the 20 different type_blocks that only exist in a reference cycle.
```

An additional perk of this change is that it now conforms `type_blocks.iloc` to the pattern of `Frame` and `Series`. I also think it's reasonable to assume/hope this change will lead to performance/memory improvements across the board.

### DIFF
--- a/static_frame/core/type_blocks.py
+++ b/static_frame/core/type_blocks.py
@@ -273,7 +273,7 @@ class TypeBlocks(ContainerOperand):
         return a
 
     @property
-    def iloc(self) -> InterfaceGetItem:
+    def iloc(self) -> InterfaceGetItem: #type: ignore
         return InterfaceGetItem(self._extract_iloc)
 
     @property #type: ignore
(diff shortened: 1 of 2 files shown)


flexatone:
Thanks for this excellent improvement. I will get it in the 0.6.22 release soon.

PR #233, opened 2020-08-12
Fix for saving frames to sqlite, when labels contain special characters.

ForeverWintr:
The case that prompted this pull request was a dash in the `label` (which becomes table name). You can see the failure by checking out db4a26a and running the `store_sqlite` tests. 

The fix is just to quote the label in generated sql.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,6 @@ dist
 
 *.egg-info
 *.so
+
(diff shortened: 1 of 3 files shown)


flexatone:
Many thanks for isolating this issue and implementing the fix.

PR #246, opened 2020-09-11
116/msgpack

MadisonAster:
Everything is working now, though it's a little inelegant. 

Notes:
* If the magic character hack is acceptable for handling arbitrary data types, a better magic character is needed.
* I used  getattr(sys.modules[foo], bar) to replace eval in the magic character hack, but maybe there's a better way?

### DIFF
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -20,5 +20,5 @@ openpyxl==3.0.0
 xarray==0.13.0
 tables==3.6.1
 pyarrow==0.16.0
-msgpack=0.6.2
-msgpack-numpy=0.4.5
+msgpack=1.0.0
+msgpack-numpy=0.4.6.1
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -20,5 +20,5 @@ openpyxl==3.0.0
 xarray==0.13.0
 tables==3.6.1
 pyarrow==0.16.0
-msgpack>=0.6.2
-msgpack-numpy>=0.4.5
+msgpack=0.6.2
+msgpack-numpy=0.4.5
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -20,5 +20,5 @@ openpyxl==3.0.0
 xarray==0.13.0
 tables==3.6.1
 pyarrow==0.16.0
-msgpack=1.0.0
-msgpack-numpy=0.4.6.1
+msgpack==1.0.0
+msgpack-numpy==0.4.6.1
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -21,4 +21,4 @@ xarray==0.13.0
 tables==3.6.1
 pyarrow==0.16.0
 msgpack==1.0.0
-msgpack-numpy==0.4.6.1
+msgpack-numpy==0.4.7
--- a/static_frame/core/frame.py
+++ b/static_frame/core/frame.py
@@ -2111,6 +2111,20 @@ class Frame(ContainerOperand):
         from datetime import datetime
         import fractions
         import sys
+        def msgpack_fixes(typ, d):
+            #Hypothesis coverage
+            if typ == 'datetime':
+                return datetime.strptime(d, '%Y %a %b %d %H:%M:%S:%f')
+            elif typ == 'date':
+                return datetime.strptime(d, '%Y %a %b %d %H:%M:%S:%f').date()
+            elif typ == 'Fraction':
+                return fractions.Fraction(d)
+            elif typ == 'int':
+                return int(d)
+            elif typ == 'ndarray':
+                return unpackb(d) #recurse unpackb
+            else:
+                return d
         def decode(obj, chain=msgpack_numpy.decode):
             if b'sf' in obj:
                 clsname = obj[b'sf']
@@ -2142,35 +2156,14 @@ class Frame(ContainerOperand):
                     blocks = unpackb(obj[b'blocks']) #recurse unpackb
                     return cls.from_blocks(blocks)
             elif b'np' in obj:
-                #Parse overridden numpy datatypes
+                #Overridden numpy datatypes
                 data = unpackb(obj[b'data']) #recurse unpackb
                 typename = obj[b'dtype'].split('[',1)[0]
-                
                 if typename in ['datetime64', 'timedelta64', '>m8', '>M8']:
                     array = np.array(data, dtype=obj[b'dtype'])
                 elif typename == 'object_':
-                    #Hypothesis coverage:
-                    
-                    result = []
-                    for (typ, d) in data:
-                        p = typ.split('.',1)[0]
-                        c = typ.split('.',1)[-1]
-                        if p == 'datetime':
-                            D = datetime.strptime(d, '%Y %a %b %d %H:%M:%S:%f')
-                            if c == 'date':
-                                D = D.date()
-                            result.append(D)
-                        elif c == 'Fraction':
-                            result.append(fractions.Fraction(d)) 
-                        elif c == 'int':
-                            result.append(int(d)) 
-                        elif c == 'ndarray':
-                            result.append(unpackb(d)) #recurse unpackb
-                        else:
-                            result.append(d)
-                    array = np.array(result, dtype=np.object_) #lots of hypothesis tests need to be 
-                                                               #explicitly cast back to object_
-                    #End Hypothesis coverage
+                    result = [msgpack_fixes(typ, d) for (typ, d) in data]
+                    array = np.array(result, dtype=np.object_)
                 array.flags.writeable = False
                 return array
             else:
@@ -5621,6 +5614,22 @@ class Frame(ContainerOperand):
         import msgpack
         import msgpack_numpy
         import datetime
+        def msgpack_fixes(a):
+            #Hypothesis coverage
+            typ = a.__class__.__name__
+            if typ in ['datetime', 'date']:
+                d = a.strftime('%Y %a %b %d %H:%M:%S:%f')
+                year = d.split(' ',1)[0].zfill(4) #datetime returns inconsistent year string for <4 digit years on some systems
+                d = year+' '+d.split(' ',1)[-1]
+                return (typ, d)
+            elif typ == 'ndarray':
+                return (typ, packb(a)) #recurse packb
+            elif typ == 'Fraction':
+                return (typ,  str(a))
+            elif typ == 'int' and len(str(a)) >=19: #msgpack-python has an overflow issue with large ints
+                return (typ, str(a))
+            else:
+                return ('', a)
         def encode(obj, chain=msgpack_numpy.encode):
             clsname = obj.__class__.__name__
             package = obj.__class__.__module__.split('.',1)[0]
@@ -5649,31 +5658,11 @@ class Frame(ContainerOperand):
             elif package == 'numpy':
                 #msgpack_numpy is breaking with these data types, overriding here
                 if isinstance(obj, np.ndarray):
-                    #Hypothesis coverage:
-
                     if obj.dtype.type == np.object_:
-                        def msgpack_fixes(a):
-                            c = a.__class__.__name__
-                            p = a.__class__.__module__.split('.',1)[0]
-                            if p == 'datetime':
-                                d = a.strftime('%Y %a %b %d %H:%M:%S:%f')
-                                year = d.split(' ',1)[0].zfill(4) #datetime returns inconsistent year string for <4 digit years on some systems
-                                d = year+' '+d.split(' ',1)[-1]
-                            elif c == 'ndarray':
-                                d = packb(a) #recurse packb
-                            elif c == 'Fraction':
-                                d = str(a)
-                            elif c == 'int' and len(str(a)) >=19: #msgpack-python has an overflow issue with large ints
-                                d = str(a)
-                            else:
-                                d = a
-                            return (p+'.'+c, d)
                         data = [msgpack_fixes(a) for a in obj]
                         return {b'np': True,
                                 b'dtype': 'object_',
                                 b'data': packb(data)} #recurse packb
-
-                    #End Hypothesis coverage
                     elif obj.dtype.type in [np.datetime64]: 
                         data = obj.astype(str)
                         return {b'np': True,
--- a/static_frame/core/frame.py
+++ b/static_frame/core/frame.py
@@ -2108,7 +2108,7 @@ class Frame(ContainerOperand):
         '''
         import msgpack
         import msgpack_numpy
-        import datetime
+        from datetime import datetime
         import fractions
         import sys
         def decode(obj, chain=msgpack_numpy.decode):
@@ -2142,46 +2142,40 @@ class Frame(ContainerOperand):
                     blocks = unpackb(obj[b'blocks']) #recurse unpackb
                     return cls.from_blocks(blocks)
             elif b'np' in obj:
+                #Parse overridden numpy datatypes
                 data = unpackb(obj[b'data']) #recurse unpackb
                 typename = obj[b'dtype'].split('[',1)[0]
+                
                 if typename in ['datetime64', 'timedelta64', '>m8', '>M8']:
                     array = np.array(data, dtype=obj[b'dtype'])
-                else:
+                elif typename == 'object_':
                     #Hypothesis coverage:
-
-                    if typename == 'datetime':
-                        data = [datetime.datetime.strptime(
-                                d, '%Y %a %b %d %H:%M:%S:%f') for d in data]
-                    elif typename == 'date':
-                        data = [datetime.datetime.strptime(
-                                d, '%Y %a %b %d %H:%M:%S:%f').date() for d in data]
-                    elif typename == 'Fraction':
-                        data = [fractions.Fraction(a) for a in data]
-                    elif typename == 'int':
-                        data = [int(n) for n in data]
-                    elif typename == 'float':
-                        data = [float(n) for n in data]
-                    elif typename == 'multitype':
-                        result = []
-                        for (typ, d) in data:
-                            p = typ.split('.',1)[0]
-                            c = typ.split('.',1)[-1]
-                            if p == 'datetime':
-                                D = datetime.datetime.strptime(d, '%Y %a %b %d %H:%M:%S:%f')
-                                if c == 'date':
-                                    D = D.date()
-                                result.append(D)
-                            elif c == 'NoneType':
-                                result.append(None)
-                            elif p == 'builtins':
-                                result.append(globals()['__builtins__'][c](d))
-                            elif p in sys.modules:
-                                result.append(getattr(sys.modules[p], c)(d))
-                            else:
-                                raise Exception('missing type!', p, c, d)
-                        data = result
-                    array = np.array(data, dtype=np.object_) #lots of hypothesis tests need to be 
-                                                             #explicitly cast back to object_
+                    
+                    result = []
+                    for (typ, d) in data:
+                        p = typ.split('.',1)[0]
+                        c = typ.split('.',1)[-1]
+                        if p == 'datetime':
+                            D = datetime.strptime(d, '%Y %a %b %d %H:%M:%S:%f')
+                            if c == 'date':
+                                D = D.date()
+                            result.append(D)
+                        elif c == 'NoneType':
+                            result.append(None)
+                        elif c == 'Fraction':
+                            result.append(fractions.Fraction(d)) 
+                        elif c == 'int':
+                            result.append(int(d)) 
+                        elif c == 'ndarray':
+                            result.append(unpackb(d)) #recurse unpackb
+                        elif p == 'builtins':
+                            result.append(globals()['__builtins__'][c](d))
+                        elif p in sys.modules:
+                            result.append(getattr(sys.modules[p], c)(d))
+                        else:
+                            raise Exception('missing type!', p, c, d)
+                    array = np.array(result, dtype=np.object_) #lots of hypothesis tests need to be 
+                                                               #explicitly cast back to object_
                     #End Hypothesis coverage
                 array.flags.writeable = False
                 return array
@@ -5662,48 +5656,33 @@ class Frame(ContainerOperand):
                 #msgpack_numpy is breaking with these data types, overriding here
                 if isinstance(obj, np.ndarray):
                     #Hypothesis coverage:
-                
+
                     if obj.dtype.type == np.object_:
-                        typeset = list(set(map(type, obj)))
-                        if len(typeset) > 1:
-                            data = []
-                            for a in obj:
-                                c = a.__class__.__name__
-                                p = a.__class__.__module__.split('.',1)[0]
-                                if p == 'datetime':
-                                    d = a.strftime('%Y %a %b %d %H:%M:%S:%f')
-                                    year = d.split(' ',1)[0].zfill(4) #datetime returns inconsistent year string for <4 digit years on some systems
-                                    d = year+' '+d.split(' ',1)[-1]
-                                elif c == 'bool':
-                                    d = int(a)
-                                else:
-                                    d = str(a)
-                                data.append((p+'.'+c, d))
-                            return {b'np': True,
-                                    b'dtype': 'multitype',
-                                    b'data': packb(data)} #recurse packb
-                        else:
-                            t = typeset[0]
-                            tname = t.__name__
-                            if tname in ['datetime', 'date']:
-                                data = []
-                                for a in obj:
-                                    d = a.strftime('%Y %a %b %d %H:%M:%S:%f')
-                                    year = d.split(' ',1)[0].zfill(4) #datetime returns inconsistent year string for <4 digit years on some systems
-                                    d = year+' '+d.split(' ',1)[-1]                              
-                                    data.append(d)
-                            elif tname == 'int': #msgpack fails on a large number case from hypothesis here
-                                data = [str(a) for a in obj]
-                            elif tname == 'Fraction':
-                                data = [str(a) for a in obj.astype(t)]
+                        result = []
+                        for a in obj:
+                            c = a.__class__.__name__
+                            p = a.__class__.__module__.split('.',1)[0]
+                            if p == 'datetime':
+                                d = a.strftime('%Y %a %b %d %H:%M:%S:%f')
+                                year = d.split(' ',1)[0].zfill(4) #datetime returns inconsistent year string for <4 digit years on some systems
+                                d = year+' '+d.split(' ',1)[-1]
+                            elif c == 'bool':
+                                d = int(a)
+                            elif c == 'ndarray':
+                                d = packb(a) #recurse packb
+                                print('recurse packb', a, d)
+                            elif c == 'Fraction':
+                                d = str(a)
+                            elif c == 'int' and len(str(a)) >=19: #msgpack-python has an overflow issue with large ints
+                                d = str(a)
                             else:
-                                data = [a for a in obj]
-                            return {b'np': True,
-                                    b'dtype': tname,
-                                    b'data': packb(data)} #recurse packb
-                    
+                                d = a
+                            result.append((p+'.'+c, d))
+                        return {b'np': True,
+                                b'dtype': 'object_',
+                                b'data': packb(result)} #recurse packb
+
                     #End Hypothesis coverage
-                    
                     elif obj.dtype.type in [np.datetime64]: 
                         data = obj.astype(str)
                         return {b'np': True,
--- a/static_frame/core/frame.py
+++ b/static_frame/core/frame.py
@@ -2186,6 +2186,8 @@ class Frame(ContainerOperand):
                 elif typename in ['complex64', '>c8']:
                     array = np.array(data, dtype=obj[b'dtype'])
                 else:
+                    #Hypothesis coverage:
+
                     if typename == 'datetime':
                         data = [datetime.datetime.strptime(
                                 d, '%Y %a %b %d %H:%M:%S:%f') for d in data]
@@ -2217,7 +2219,9 @@ class Frame(ContainerOperand):
                             else:
                                 raise Exception('missing type!', p, c, d)
                         data = result
-                    array = np.array(data, dtype=np.object_)
+                    array = np.array(data, dtype=np.object_) #lots of hypothesis tests need to be 
+                                                             #explicitly cast back to object_
+                    #End Hypothesis coverage
                 array.flags.writeable = False
                 return array
             else:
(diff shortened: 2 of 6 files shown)


static_frame/core/frame.py
```diff
@@ -5968,6 +6013,27 @@ def to_latex(self,
...
+        import msgpack
+        import msgpack_numpy
+        def cast_msgpack(input):
+            try:
+                data = msgpack.packb(input, default=msgpack_numpy.encode) #try coercing standard and numpy datatypes
+            except TypeError:
+                #This inserts a magic character with any arbitrary datatype that msgpack_numpy doesn't support
+                #TODO: Is this attempting to handle too much? Is there a better way?
+                clsname = '>'+input[0].__class__.__module__+'.'+input[0].__class__.__name__                
+                data = msgpack.packb([clsname]+[a.__str__() for a in input], use_bin_type=True) #else cast to string
```
flexatone, line 6029: Encoding the string form of each data element seems a bit inefficient: is there no other option? And, on import, will we get back the types we originally had? If I have an `Index` of integers, will we get back integers on import?

static_frame/core/frame.py
```diff
@@ -5968,6 +6013,27 @@ def to_latex(self,
...
+            return data
+        return cast_msgpack({
+            '_index' : cast_msgpack([index for index in self._index]),
+            '_columns' : cast_msgpack([column for column in self._columns]),
```
flexatone, line 6033: For 1D `Index`, you want to get the NumPy array via `.values` instead of iterating through the `Index` and creating a new list. For 2D `IndexHierarchy`, you want to use the `._blocks` attribute, which is a `TypeBlocks` instance. You can get the list of contained arrays with `._blocks._blocks`.

We also need to capture the `name` of the `Index`.

static_frame/core/frame.py
```diff
@@ -5968,6 +6013,27 @@ def to_latex(self,
...
+            '_name' : cast_msgpack(self._name),
+            '_blocks' : [cast_msgpack([value[0] for value in block.values]) for block in self.transpose()._blocks],
```
flexatone, line 6035: Encoding at the block level is the right approach here, but why are we transposing first? Transposing may introduce undesirable type coercion. I would just try to encode each np.array in `self._blocks._blocks` (that is the underlying list of np.arrays).

static_frame/test/unit/test_frame.py
```diff
@@ -6160,9 +6160,32 @@ def test_frame_to_latex_a(self) -> None:
...
+        records = (
+                (2, 'a', False),
+                (3, 'b', False),
+                )
+        f1 = Frame.from_records(records,
+                columns=('r', 's', 't'),
+                index=('w', 'x'))
+        post = f1.to_msgpack()
+        msg = b'\x84\xa6_index\xc4\x05\x92\xa1w\xa1x\xa8_columns\xc4\x07\x93\xa1r\xa1s\xa1t\xa5_name\xc4\x01\xc0\xa7_blocks\x92\xc4\x05\x93\x02\xa1a\xc2\xc4\x05\x93\x03\xa1b\xc2'
+        self.assertEqual(post, msg)
```
flexatone, line 6174: Ultimately you will want to test that, given `f2 = Frame.from_msgpack(msg)`, `f1.equals(f2, compare_name=True, compare_dtype=True, compare_class=True)`.

static_frame/core/frame.py
```diff
@@ -1393,6 +1393,51 @@ def from_json(cls,
...
+        unpacked = uncast_msgpack(msgpack_data)
+        for key in ['_name','_index','_columns']:
+            unpacked[key] = uncast_msgpack(unpacked[key])
+        for i in range(len(unpacked['_blocks'])):
+            unpacked['_blocks'][i] = uncast_msgpack(unpacked['_blocks'][i])
+        
+        return cls.from_records(unpacked['_blocks'],
+                columns=unpacked['_columns'],
+                index=unpacked['_index'],
+                name=unpacked['_name'])
```
flexatone, line 1439: I think you are using `from_records` here because you transposed the `Frame`, below. Better to not transpose, and create a `TypeBlocks` instance directly from `TypeBlocks.from_blocks`, then feed that to `Frame.__init__` constructor setting `own_data=True`.

static_frame/test/unit/test_frame.py
```diff
@@ -6160,9 +6160,32 @@ def test_frame_to_latex_a(self) -> None:
...
-
+    
+    def test_frame_to_msgpack_a(self) -> None:
+        records = (
+                (2, 'a', False),
+                (3, 'b', False),
+                )
+        f1 = Frame.from_records(records,
+                columns=('r', 's', 't'),
+                index=('w', 'x'))
```
flexatone, line 6171: If we have integer labels here will they come back as integers or strings?

static_frame/core/frame.py
```diff
@@ -1393,6 +1393,51 @@ def from_json(cls,
...
+            if isinstance(data[0], str) and data[0][0] == '>': #if we detect the magic character, evaluate class
+                import sys
+                evaldata = []
+                clsname = data[0][1:]
+                m, c = clsname.split('.',1)
+                cls = getattr(sys.modules[m], c)
+                #cls = eval(clsname) #replacement code above, leaving this in so that it's obvious what's going on.
+                
+                for d in data[1:]:
+                    evaldata.append(cls.__new__(cls, d))
```
flexatone, line 1427: Why call `cls.__new__` here instead of just `cls(d)`?

flexatone:
Many thanks for this PR! Many of my comments are about trying to preserve the underlying NumPy arrays, which relies to some degree on undocumented interfaces, so I know that will take some exploration.

static_frame/core/frame.py
```diff
@@ -5968,6 +6013,27 @@ def to_latex(self,
...
+        '''
+        Return a msgpack.
+        '''
+        import msgpack
+        import msgpack_numpy
+        def cast_msgpack(input):
+            try:
+                data = msgpack.packb(input, default=msgpack_numpy.encode) #try coercing standard and numpy datatypes
+            except TypeError:
+                #This inserts a magic character with any arbitrary datatype that msgpack_numpy doesn't support
```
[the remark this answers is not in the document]
MadisonAster: Specifically np.datetime64. which I covered in test_frame_from_msgpack_b, it would be beneficial to convert that specifically to an int and a custom datatype, however I was guessing there might be many cases like this. I also figured msgpack_numpy developers would have already added a clean way to do it if it was easy.

I added the hack to ensure we would get the same data types when running from_msgpack, however if the goal is to share the msgpack with other applications, using a plain string as the fallback would be more desirable.

static_frame/core/frame.py
MadisonAster: Definitely inefficient. This is the last resort fallback for anything msgpack_numpy can't handle.

static_frame/core/frame.py
MadisonAster: Ok, great. I figured I was missing some parts here, I'll work on understanding those use cases.

static_frame/core/frame.py
MadisonAster: I used transpose out of my ignorance of static_frame's structure, self._blocks._blocks sounds like a much better approach

static_frame/test/unit/test_frame.py
MadisonAster: thanks! I was having trouble finding an example of another test that worked like this to copy syntax from.

static_frame/core/frame.py
MadisonAster: The other way around actually, I used transpose to make this work. Still bad :P. I found TypeBlocks, but didn't figure from_blocks yet.

static_frame/test/unit/test_frame.py
MadisonAster: I'll add more test coverage. Also realized last night I forgot to add tests for things like np.float_64 and np.int64, which was the whole reason for involving msgpack_numpy.

static_frame/core/frame.py
MadisonAster: Yeah that's better! Sorry my brain folded over on itself one too many times there.

static_frame/core/frame.py
MadisonAster: The better option is to build custom mspack datatypes for every unsupported datatype that we need, but I think this is the best fallback option for making sure things still work when library users start injecting their own foreign data types.

static_frame/core/frame.py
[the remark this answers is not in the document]
MadisonAster: btw here is the example use case using yfinance that I built to play around with that led me to discovering np.datetime64 was unsupported:

[KungFu/Python/DataStructures/Array/StaticFrameFromSymbol.py](https://github.com/MadisonAster/KungFu/blob/master/Python/DataStructures/Arrays/StaticFrameFromSymbol.py)

[KungFu/Python/DataStructures/Binary/MessagePack.py](https://github.com/MadisonAster/KungFu/blob/master/Python/DataStructures/Binary/MessagePack.py)

static_frame/core/frame.py
MadisonAster: updated functions to use _blocks._blocks. Undesirable coercion was occurring in test_frame_from_msgpack_c, everything was turned to float64 from the transposition.

static_frame/test/unit/test_frame.py
MadisonAster: tests now using equals method for comparison.

static_frame/core/frame.py
MadisonAster: from_msgpack now returns instance using TypeBlocks.from_blocks setting own_data=True

static_frame/test/unit/test_frame.py
MadisonAster: Better tests added covering:
int
np.float64
np.short
np.int64
np.datetime64

see :
test_frame_from_msgpack_b
test_frame_from_msgpack_c

static_frame/core/frame.py
MadisonAster: cleaned up this section

static_frame/core/frame.py
MadisonAster: now using .values for 1D arrays instead of repacking

MadisonAster: preserving index name now

static_frame/core/frame.py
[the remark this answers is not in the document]
flexatone: Interesting: so is it only for np.datetime64 types that we need the "Magic Character Hack"? If that is the case perhaps we can just identify that type and perform a special encoding/decoding. I think it is the case that all np.datetime64 can be represented in integers, and be converted back to np.datetime64 as long as you know the dtype. np.datetime64 could then be encoded as a pair of int array and string dtype representation.

```python
>>> a1 = np.array(['2010', '2013'], dtype=np.datetime64)
>>> a1
array(['2010', '2013'], dtype='datetime64[Y]')
>>> a1.astype(int)
array([40, 43])
>>> a1.dtype.str
'<M8[Y]'
>>> np.array(a1.astype(int), dtype=a1.dtype.str)
array(['2010', '2013'], dtype='datetime64[Y]')
```

static_frame/core/frame.py
flexatone: Perhaps handling datetime64 explicitly will remove the need for this.

static_frame/test/unit/test_frame.py
flexatone: Very nice expansion of the test coverage! When you want to go even further, you can try implementing a property test with Hypothesis (https://hypothesis.readthedocs.io). SF has a library of Hypothesis strategies that "randomly" generate incredibly diverse ``Frame`` and ``Series`` for testing. Take a look at `test.property.test_frame.test_frame_to_xlsx` to see an example. This will inevitable shake out edge cases and types you were not expecting.

static_frame/core/frame.py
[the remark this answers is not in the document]
MadisonAster: it applies to timedelta64, and datetime64. I pushed new commits last night that take care of the same thing with dtype, but still using the magic character to pass the information across. It's only working for 1D arrays like index and columns at the moment, and I need to get it working for column types as well. See test_frame_from_msgpack_d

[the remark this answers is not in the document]
MadisonAster: We can almost certainly pass the type information for this with a single character, instead of using the hack. I need to figure that part of msgpack out still.

static_frame/test/unit/test_frame.py
MadisonAster: I definitely want to read up on that. This is the first time I've ever encountered that library. So many fun new concepts ^-^

flexatone:
Great progress on this! Let me summarize what I see as the next steps.

1. See if we can avoid the magic character by specifically handling np.datetime64, encoding them as pairs of integer array, dtype string.
2. I see you are capturing `Frame.index.name`; we also need `Frame.columns.name`.
3. We need to encode the class of `.index` and `.columns`, as there are specialized subclasses such as `IndexDate`. (`IndexHierarchy` is not subclass of `Index`, but of `IndexBase`.)
4. We need to handle `IndexHierarchy` found on either  `.index` or `.columns`; for those, you will want to encode the `_blocks._blocks` arrays.
5.  `IndexHiearchy` (`IndexHierarchyGO` too) is lazy in its creation of ._blocks. Thus, you will need to reach into its private interface and do something like the following to force updating the `_blocks`.
```python
if ih._recache:
    ih._update_array_cache() 
```
6. Implement a property test to test more diverse indices, dtypes, and block structures.

MadisonAster:
Sounds great! I'll start looking into these items. Thanks for taking the time to review!

static_frame/core/frame.py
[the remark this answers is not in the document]
flexatone: I see. Encoding `timedelta64` as a pair of int array, dtype string might also be possible.

flexatone:
Another thing: when bringing back arrays from msgpack, every array should be made immutable by setting `array.flags.writeable = False`. This will permit SF objects to compose those arrays without making a copy.

MadisonAster:
Added: columns.name
Added: array.flags.writeable = False.

Refactored the code to work more similarly to msgpack_numpy. I should have investigated the source for that sooner. All current tests are passing. I think it looks a lot cleaner now.

Working on understanding Hierarchies better tomorrow so I can complete items 3-6.

MadisonAster:
👍 [The Promise of Hierarchical Indices for Data Beyond 2 Dimensions | PyData LA 2019](https://youtu.be/xX8tXSNDpmE)

MadisonAster:
Refactored again, part-way through getting IndexHierarchy in test f.

static_frame/core/frame.py
```diff
@@ -5968,6 +6021,55 @@ def to_latex(self,
...
+        Return a msgpack.
+        '''
+        import msgpack
+        import msgpack_numpy
+        def encode(obj, chain=msgpack_numpy.encode):
+            clsname = obj.__class__.__name__
+            package = obj.__class__.__module__.split('.',1)[0]
+            print('package, clsname', package, clsname)
+            if package == 'static_frame':
+                if clsname in ['Index', 'IndexBase']:
```
flexatone, line 6035: Rather than match to `clsname`, can you just use the `obj` reference and match to `isinstance(obj, IndexBase)`? That way you will match to all of the derived `Index` objects, such as `IndexGO` and `IndexNanosecond`.

MadisonAster: 100% agree. I deviated from using isinstance for the moment so I could explicitly write out all of the classes as I'm learning more of the structure. Wanted to be sure I had everything covered explicitly, then go back to covering it implicitly with isinstance when I tidy up.

static_frame/core/frame.py
```diff
@@ -5968,6 +6021,55 @@ def to_latex(self,
...
+                        obj._update_array_cache()
+                    return {b'sf':clsname,
+                            b'name':obj.name,
+                            b'data':packb(obj._blocks._blocks), #recursively call packb to nest msgpacks within
+                            b'levels':packb(obj._levels)} #recursively call packb to nest msgpacks within
+                elif clsname in ['IndexLevel']:
+                    return {b'sf':clsname,
+                            b'index':packb(obj.index), #recursively call packb to nest msgpacks within
+                            b'_depth':obj._depth,
+                            b'_length':obj._length}
```
flexatone, line 6050: This is a nice approach to handle the `IndexLevel` instance in an `IndexHiearchy`, but I am not certain if we should store it: it is a complicated, nested structure that is the "tree" form of what is defined in the `TypeBlocks` composed within `IndexHierarchy`. Said another way, you can get by with only storing either the `TypeBlocks` or the `IndexLevel`. If we store both, the memory footprint will be bigger, but instantiation on decoding will be faster. Because `IndexLevel` is nested, it might take longer to encode.

static_frame/core/frame.py
```diff
@@ -5968,6 +6021,55 @@ def to_latex(self,
...
+                            b'columns':packb(obj.columns)} #recursively call packb to nest msgpacks within
+                elif clsname in ['IndexDate']:
+                    return {b'sf':clsname,
+                           }
+            if package == 'numpy':
+                if isinstance(obj, np.ndarray):
+                    if isinstance(obj[0], np.datetime64) or isinstance(obj[0], np.timedelta64):
+                        #TODO: Couldn't find an attribute for unit, just splitting it from the name for now
+                        unit = str(obj[0].dtype).split('[',1)[-1].split(']',1)[0]
+                        data = [[int(a.astype(int)), unit] for a in obj]
```
flexatone, line 6065: Can we just convert the entire array at once, i.e., `obj.astype(int)`, and then store the unit in a different field of your dictionary? It will be a lot faster than doing the list comprehension.

static_frame/core/frame.py
```diff
@@ -5968,6 +6021,55 @@ def to_latex(self,
...
+                            b'index':packb(obj.index), #recursively call packb to nest msgpacks within
+                            b'columns':packb(obj.columns)} #recursively call packb to nest msgpacks within
+                elif clsname in ['IndexDate']:
+                    return {b'sf':clsname,
+                           }
+            if package == 'numpy':
+                if isinstance(obj, np.ndarray):
+                    if isinstance(obj[0], np.datetime64) or isinstance(obj[0], np.timedelta64):
+                        #TODO: Couldn't find an attribute for unit, just splitting it from the name for now
+                        unit = str(obj[0].dtype).split('[',1)[-1].split(']',1)[0]
```
flexatone, line 6064: You should be able to just use the `str` of the dtype here.

flexatone:
```>>> a1 = np.array(('2020', '20201'), dtype=np.datetime64)
>>> str(a1.dtype)
'datetime64[Y]'
```

static_frame/core/frame.py
```diff
@@ -5968,6 +6021,55 @@ def to_latex(self,
...
+                            b'name':obj.name,
+                            b'data':packb(obj._blocks._blocks), #recursively call packb to nest msgpacks within
+                            b'index':packb(obj.index), #recursively call packb to nest msgpacks within
+                            b'columns':packb(obj.columns)} #recursively call packb to nest msgpacks within
+                elif clsname in ['IndexDate']:
+                    return {b'sf':clsname,
+                           }
+            if package == 'numpy':
+                if isinstance(obj, np.ndarray):
+                    if isinstance(obj[0], np.datetime64) or isinstance(obj[0], np.timedelta64):
```
flexatone, line 6062: Rather than looking at the type of the element in the array, you can simply look at the `kind` of the `dtype` attr, 
```
>>> np.array(('2020', '20201'), dtype=np.datetime64).dtype.kind
'M'
```
You want to handle here kinds 'M' (datetime64) and 'm' (timedelta64)

static_frame/core/frame.py
```diff
@@ -1393,6 +1393,59 @@ def from_json(cls,
...
+                            columns = unpackb(obj[b'columns']),
+                            own_data=True)
+                elif clsname in ['Index', 'IndexBase']:
+                    return cls(data, name = obj[b'name'])
+                elif clsname in ['IndexHierarchy', 'IndexHierarchyGO']:
+                    return cls(
+                            levels=unpackb(obj[b'levels']),
+                            name=obj[b'name'],
+                            blocks=TypeBlocks.from_blocks(data),
+                            own_blocks=True)
```
flexatone, line 1435: If you go with just encoding the `TypeBlocks` of the `IndexHiearchy`, you can use the private constructor `IndexHierarchy._from_type_blocks` to reinstantiate the instance.

static_frame/core/frame.py
```diff
@@ -1393,6 +1393,59 @@ def from_json(cls,
...
+                elif clsname in ['IndexLevel']:
+                    return cls(unpackb(obj[b'index']))
+            elif b'np' in obj:
+                data = msgpack.unpackb(obj[b'data'])
+                cls = getattr(np, obj[b'np'])
+                array = np.array([cls(d, unit) for d, unit in data])
```
flexatone, line 1441: Rather than doing the list comprehension and applying `unit` to each item, if we store an `int` array and a single `unit`, we should be able to simply do `data.astype(unit)`, which will be very fast.

flexatone:
This is an excellent evolution from where we started. I had been looking at the source for `msgpack_numpy` and was going to suggest bundling the objects in dictionaries, as is done there, but you have gone further and really extended `msgpack_numpy` to handle SF objects, which seems like the ideal approach!

A few summary observations

1. Try to deal with the datetime64 arrays as arrays and avoid converting the to lists of objects.
2. Perhaps we do not need to store the `IndexLevel` within `IndexHierarchy`; there are performance trade-offs to consider (both positive and negative). If we do not store `IndexLevel`, we will need to store the `index_constructors` to pass to `IndexHierarchy._from_type_blocks`. You can get this via `IndexHierarchy.index_types`.

static_frame/core/frame.py
MadisonAster: using _from_type_blocks now. got test f working with this!

static_frame/core/frame.py
MadisonAster:
```
#if obj.dtype.kind in ['M', 'm']:
 if obj.dtype.type in [np.datetime64, np.timedelta64]:
```
I spent a bunch of time today playing with dtype to understand it better. I dtype.type makes the code more clear than dtype.kind

static_frame/core/frame.py
MadisonAster: This worked, and helped me clean up from_msgpack!

static_frame/core/frame.py
MadisonAster: Doing that now using the key b'dtype'

static_frame/core/frame.py
MadisonAster: Removed the IndexLevel and started using _from_type_blocks per your suggestion below.

static_frame/core/frame.py
MadisonAster: I got it working like this:
`np.array(data, dtype=obj[b'dtype'])`
I think this should be the same performance right?

MadisonAster:
Thanks for all your help with this Chris! I've added a hypothesis test, and it helped me identify and resolve the class conflict you mentioned would occur. However it seems like msgpack_numpy is chocking when given a particular scenario involving complex64 numbers. I'm not sure if I'm reading it correctly but it may be something we would need to fix in their library. I've left some debug output in the function if you want to run the test. 

test name is:  test_frame_to_msgpack

static_frame/core/frame.py
flexatone: That is a fine approach. I have typically used dtype.kind because the string comparisons are fast and light-weight, but comparing the type is fine.

flexatone (commented):
Great progress on this! For now, you can adjust the hypothesis strategy dtype to use simpler types and avoid the complex number issue. If you are finding that `msgpack_numpy` does not support some complex numbers, we should isolate it and create an issue for that project.

Other than than, I think this is really close to merging. The last things I see necessary:

1. Encode and decode index types for `IndexHierarchy`.
2. Update requirements-test.txt, requirements-extras.txt with the new msgpack pack requirements.

static_frame/core/frame.py
```diff
@@ -2098,6 +2098,88 @@ def from_parquet(cls,
...
+                        *IndexBase            #CONFLICT: IndexBase must come after IndexHierarchy.
+                        IndexHierarchy
+                        IndexHierarchyGO
+                    '''
+                    blocks = unpackb(obj[b'blocks']) #recurse unpackb
+                    return cls._from_type_blocks(
+                            blocks=blocks,
+                            name=obj[b'name'],
+                            own_blocks=True)
+                elif issubclass(cls, IndexBase):
```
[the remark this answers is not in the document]
MadisonAster: Changed from IndexBase to Index. 

I think I understand the problem fully now, and the intention of the IndexBase class. Since as a base class IndexBase is never called directly, we don't need to worry about a direct instance of it being an attribute of the data frame. Good thinking, keeps things clean.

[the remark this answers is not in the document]
MadisonAster: also updated requirements files

static_frame/core/frame.py
```diff
@@ -5534,6 +5616,87 @@ def to_parquet(self,
...
+                        *IndexBase            #CONFLICT: IndexBase must come after IndexHierarchy.
+                        IndexHierarchy
+                        IndexHierarchyGO
+                    '''
+                    if obj._recache:
+                        obj._update_array_cache()
+                    return {b'sf':clsname,
+                            b'name':obj.name,
+                            b'blocks':packb(obj._blocks)} #recurse packb
+                elif isinstance(obj, IndexBase):
```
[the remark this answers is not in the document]
MadisonAster: changed from IndexBase to Index

static_frame/core/frame.py
```diff
@@ -5534,6 +5616,87 @@ def to_parquet(self,
...
+                        *ContainerOperand
+                        *IndexBase            #CONFLICT: IndexBase must come after IndexHierarchy.
+                        IndexHierarchy
+                        IndexHierarchyGO
+                    '''
+                    if obj._recache:
+                        obj._update_array_cache()
+                    return {b'sf':clsname,
+                            b'name':obj.name,
+                            b'blocks':packb(obj._blocks)} #recurse packb
```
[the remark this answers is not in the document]
MadisonAster: added support for index_constructors via values.tolist()

static_frame/core/frame.py
```diff
@@ -2098,6 +2098,88 @@ def from_parquet(cls,
...
+                        *ContainerOperand
+                        *IndexBase            #CONFLICT: IndexBase must come after IndexHierarchy.
+                        IndexHierarchy
+                        IndexHierarchyGO
+                    '''
+                    blocks = unpackb(obj[b'blocks']) #recurse unpackb
+                    return cls._from_type_blocks(
+                            blocks=blocks,
+                            name=obj[b'name'],
+                            own_blocks=True)
```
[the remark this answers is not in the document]
MadisonAster: added support for parsing index_constructors via globals()[clsname].

MadisonAster:
Hey Chris, I’ve got the hypothesis tests passing with `DTGroup.CORE` now. I’ve added comments to the code noting where the hypothesis coverage is. If you switch to `DTGroup.ALL` there is a precision error that’s produced in a timedelta64 case. Not sure if it’s relevant or not. I’ve left some print statements in the test so you can take a look if you want.

Plenty to clean up still, but let me know what you think. I think most of this new logic ought to be reformatted and pushed back to msgpack_numpy instead of living here.

Also travis is complaining, but I checked the log and I don’t think that’s my stuff.

Hope you’re having a fun weekend!

MadisonAster:
CI is passing now.

static_frame/core/frame.py
```diff
@@ -5534,6 +5628,103 @@ def to_parquet(self,
...
+                                    data.append(d)
+                            elif tname == 'int': #msgpack fails on a large number case from hypothesis here
+                                data = [str(a) for a in obj]
+                            elif tname == 'Fraction':
+                                data = [str(a) for a in obj.astype(t)]
+                            else:
+                                data = [a for a in obj]
+                            return {b'np': True,
+                                    b'dtype': tname,
+                                    b'data': packb(data)} #recurse packb
```
flexatone, line 5705: We can assume arrays with object dtype always have hetergenous types, and thus only keep the branch where `len(typeset) > 1`: that we do not have to duplicate the handling for all types here.

But even then it seems like we are doing a lot of extra work here for object arrays. Can we simply use `array.tolist()` and let msgpack handle the list as it would for any other list of objects? This might be tricky, on the decode side, to make sure we re-create the array.

flexatone (commented):
Many thanks for these updates. 

Hypothesis shakes out a lot of corner cases: we want to try to handle them, but if it makes our implementation too cumbersome, we can skip those corner cases for now.

If `msgpack_numpy` does not support object arrays at all, it might be nice to contribute to that library. In the mean time, if we can convert the array to a list and pass it on to `msgpack`, that seems like a practical way to handle object arrays.

requirements-test.txt
```diff
@@ -20,3 +20,5 @@ openpyxl==3.0.0
 xarray==0.13.0
 tables==3.6.1
 pyarrow==0.16.0
+msgpack>=0.6.2
+msgpack-numpy>=0.4.5
```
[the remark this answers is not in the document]
MadisonAster: fixed

static_frame/core/frame.py
MadisonAster: I refactored the code to remove the excess logic branch, and I've moved the fix into its own function. I couldn't come up with a way to effectively use `arra.tolist()` to avoid the extra code, as hypothesis includes tests with nested ndarrays. However I have minimized the fix as much as possible.
[https://github.com/InvestmentSystems/static-frame/blob/795143b92a8de9fec749a1f90dd48ef7d504f789/static_frame/core/frame.py#L5626-L5647](https://github.com/InvestmentSystems/static-frame/blob/795143b92a8de9fec749a1f90dd48ef7d504f789/static_frame/core/frame.py#L5626-L5647)

I didn't want to give the same treatment to the static_frame objects, because then the convenience of simply calling up the class objects from `globals()[clsname]` would be destroyed, and that would create extra maintenance for any new sub-classes added to static-frame.

flexatone: Sure, this is good for now. We might be able to revisit this if `msgpack_numpy` improves support for object arrays in the future.

flexatone:
Many thanks for your contributions with this PR!

flexatone:
I have moved things around a little bit to bring together, in one class, the special handling you created for the types that `msgpack` does not support. I also made a few small changes for style and typing. Overall, very nice work with this. This will be part of the 0.6.35 release, to be out soon.

https://github.com/InvestmentSystems/static-frame/commit/7e461d12a014b6a96cda872c5358d7a989bbc7e7

MadisonAster:
I'm honored to be of help. Thank you so much for all your assistance along the way!

I'm working on trying out static frame in conjunction with jupyter lab today, so I can play with some examples that will help me better understand the overall structure. Do you know of any good public examples of ipynb files using static frame that I can dig through?

flexatone:
Sorry for the delay! There might not be many public examples using StaticFrame; but you might find an interesting example using Pandas and convert it to StaticFrame.

PR #259, opened 2020-10-02
Makes from_pandas failure messages more specific.

chaburkland:


flexatone:
Many thanks for these improvements.

PR #283, opened 2020-12-01
Test using Python 3.8.

brandtbucher:
This PR:

- Sets `CFLAGS` to ignore the compiler warnings that were failing Pandas builds for Python 3.8 using Clang (meaning we can test on Python 3.8 / macOS again).
- Adds caching for pip's built wheels, which can reduce the overhead of each run by up to 12 minutes.
- Removes duplicate jobs. Currently, all PR commits run each job *twice*, which takes forever.

In short, it makes CI better and faster. 🙂

### DIFF
--- a/.github/workflows/quality.yml
+++ b/.github/workflows/quality.yml
@@ -1,6 +1,10 @@
 name: Quality
 
-on: [push, pull_request]
+on:
+  push:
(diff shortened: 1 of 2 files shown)


flexatone:

PR #298, opened 2021-01-06
Adds `columns_select` argument to `Frame.from_sql` PR

### LINKED ISSUE #297: Add `columns_select` as optional argument to `Frame.from_sql`
Allow users to filter out columns after the sql is executed.

This is necessary since some executable sql code (i.e. stored procedures) are not easily modifiable with something akin to:
`SELECT A, B, ... FROM ...`

This will mirror the same behavior as the argument `columns_select` in `Frame.from_parquet`.

chaburkland:
Closes #297.


static_frame/core/index_hierarchy.py
```diff
@@ -368,7 +368,10 @@ def to_label(label: str) -> tp.Tuple[tp.Hashable, ...]:
             if len(parts) <= 1:
                 raise RuntimeError(f'Could not not parse more than one label from delimited string: {label}')
 
-            return tuple(literal_eval(p) for p in parts)
+            try:
+                return tuple(literal_eval(p) for p in parts)
+            except ValueError as e:
+                raise ValueError('A label is malformed. This is most likely due to not quoting a string label') from e
```
flexatone, line 374: This is a nice improvement!

static_frame/core/frame.py
```diff
@@ -1514,21 +1516,36 @@ def from_sql(cls,
...
             index = None
-            row_gen_final = lambda: row_gen
+            row_gen = lambda: sql_iterable
+
+        if columns_select:
+            columns_select = set(columns_select)
+
+            sel = columns.isin(columns_select)
+            # desired_cols = columns.loc_to_iloc(sel) # Currently doesn't work for IH...
+            desired_cols = np.where(sel)[0]
```
flexatone, line 1540: I see that a Boolean array given to `loc_to_iloc` does not give the expected result; let me see if I can fix that, as having to use `np.where` here seems undesirable. I will try to push a fix for this later today.

static_frame/core/frame.py
```diff
@@ -1493,17 +1495,17 @@ def from_sql(cls,
...
         columns = None
         if columns_depth == 1:
-            columns = cls._COLUMNS_CONSTRUCTOR(b[0] for b in row_gen.description[index_depth:])
+            columns = cls._COLUMNS_CONSTRUCTOR(b[0] for b in sql_iterable.description[index_depth:])
             own_columns = True
         elif columns_depth > 1:
             # use IH: get via static attr of columns const
             constructor = cls._COLUMNS_HIERARCHY_CONSTRUCTOR.from_labels_delimited
-            labels = (b[0] for b in row_gen.description[index_depth:])
+            labels = (b[0] for b in sql_iterable.description[index_depth:])
```
flexatone, line 1508: I wonder if there is a way we can use `columns_select` here, in the creation of columns, rather than having to trim columns later after creating columns here. That way we save creating potentially huge `Index` if the user only is requesting a single column. We might use `sql_iterable.description` and `columns_select` to also determine the appropriate indexes to extract with `itemgetter`, below, saving us from having to use the columns to determine that.

static_frame/core/frame.py
```diff
@@ -1514,21 +1516,36 @@ def from_sql(cls,
...
+        if columns_select:
+            columns_select = set(columns_select)
+
+            sel = columns.isin(columns_select)
+            # desired_cols = columns.loc_to_iloc(sel) # Currently doesn't work for IH...
+            desired_cols = np.where(sel)[0]
+
+            def row_gen_final() -> tp.Iterator[tp.Sequence[tp.Any]]:
+                for row in row_gen():
+                    yield itemgetter(*desired_cols)(row)
```
flexatone, line 1544: I like this use of `itemgetter`! However, can we call `itemgetter(*desired_cols)` once, before the `for` loop, and then simply re-use it for each row?

flexatone (changes requested):
Many thanks for this useful PR.

static_frame/core/frame.py
chaburkland: Good point. I'll optimize this.

static_frame/core/frame.py
chaburkland: I didn't think of that - will change.

static_frame/core/frame.py
```diff
@@ -1493,42 +1495,66 @@ def from_sql(cls,
...
                         yield row[index_depth:]
         else:
             index = None
-            row_gen_final = lambda: row_gen
+            row_gen = lambda: sql_iterable
+
+        if columns_select:
+            row_gen_final = lambda: (filter_row(row) for row in row_gen())
+        else:
+            row_gen_final = row_gen
```
flexatone, line 1557: Rather then pack the iterators behind `lambda`, maybe just call `row_gen` here (line 1557) and then do not call it in line 1561. That will save some unnecessary lambda calls, as well as assigning the lambda to a name.

static_frame/core/frame.py
```diff
@@ -1493,42 +1495,66 @@ def from_sql(cls,
...
+                if columns_select:
+                    # Have to wait until columns is already constructed in order to filter, since
+                    # `from_labels_delimited` is implementing specific logic to convert the SQL columns
+                    # into an IndexHierarchy, thus making filtering on the raw labels effectively impossible
+                    columns_select = set(columns_select)
+                    sel = columns.isin(columns_select)
+                    columns = columns.loc[sel]
+
+                    selector = itemgetter(*np.where(sel)[0])
+                    filter_row = lambda row: selector(row)
```
flexatone, line 1529: I have updated `IndexHierarhcy.loc_to_iloc` to support Booleans, so you can now just do `itemgetter(*columns.loc_to_iloc(columns_select))`

static_frame/core/frame.py
```diff
@@ -1493,42 +1495,66 @@ def from_sql(cls,
...
+        own_columns = False
+        filter_row = lambda row: row
+
+        if columns_depth >= 1:
+            labels = (col for (col, *_) in sql_iterable.description[index_depth:])
             own_columns = True
 
+            if columns_depth == 1:
+                if columns_select:
+                    indices, labels = map(tuple, zip(*((i, label) for (i, label) in enumerate(labels) if label in columns_select)))
```
flexatone, line 1510: As we are using `in` on a list, maybe better to maybe convert `columns_select` to a set before this move?

static_frame/core/frame.py
flexatone: You can also use the result of `loc_to_iloc` to derive the reduced columns with `columns.iloc`, reducing the number of lookups/comparisons to `columns_select`

flexatone (changes requested):
Nice changes; a few more improvements might be possible.

flexatone (approved):

PR #308, opened 2021-01-21
Use cursor object to fetch data.

gdevanla:
We need to assume that the connection object passed to from_sql does
not implement `execute` method, but implements the `cursor` method
specified by PEP 249.

### DIFF
--- a/static_frame/core/frame.py
+++ b/static_frame/core/frame.py
@@ -1498,7 +1498,18 @@ class Frame(ContainerOperand):
             {name}
             {consolidate_blocks}
         '''
-        sql_iterable = connection.execute(query)
+
+        # We cannot assume the cursor object returned by DBAPI Connection to have a context manager
+        cursor = None
+        try:
+            cursor = connection.cursor()
+            cursor.execute(query)
+            cursor_description = cursor.description
+            rows = cursor.fetchall()
+        finally:
+            if cursor:
+                cursor.close()
+
         columns = None
         own_columns = False
 
@@ -1511,7 +1522,7 @@ class Frame(ContainerOperand):
 
         if columns_depth >= 1 or columns_select:
             # always need to derive labels if using columns_select
-            labels = (col for (col, *_) in sql_iterable.description[index_depth:])
+            labels = (col for (col, *_) in cursor_description[index_depth:])
 
         if columns_depth <= 1 and columns_select:
             iloc_sel, labels = zip(*(
@@ -1543,7 +1554,7 @@ class Frame(ContainerOperand):
                 index_constructor = Index
 
                 def row_gen() -> tp.Iterator[tp.Sequence[tp.Any]]:
-                    for row in sql_iterable:
+                    for row in rows:
                         index.append(row[0])
                         yield row[1:]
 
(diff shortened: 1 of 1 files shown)


static_frame/core/frame.py
```diff
@@ -1498,7 +1498,18 @@ def from_sql(cls,
...
         '''
-        sql_iterable = connection.execute(query)
+
+        # We cannot assume the cursor object returned by DBAPI Connection to have a context manager
+        cursor = None
+        try:
+            cursor = connection.cursor()
+            cursor.execute(query)
+            cursor_description = cursor.description
+            rows = cursor.fetchall()
```
flexatone, line 1508: Is there a way to do this without using `fetchall()`? Doing so greedily grabs all data, while the previous implementation lazily read through rows, potentially saving creation of a large list.

flexatone (changes requested):
In general, this looks like a good change. I have one in-line request. Also: can we add at least one test to make sure we have coverage for the new branches?

static_frame/core/frame.py
gdevanla: I think the only way to continue to keep this process lazy, is to wrap the entire function in try/finally.

Couple other observations on the current implementation:

1.  `connection.execute()`, implicity creates a cursor. Which means the current implementation was not closing the cursor before returning.  This might not impact general uses cases, but for long living connections, this cursor object will always be open.

2.  I see that index=[], is being mutated inside the row_gen() as a closure variable.  There is a risk of reading this `index` variable before it is fully populated in the function it is being passed into.  Is that a conscious design decision?

flexatone: At this time, `Frame.from_records` has to take a `list` of a generator anyway, so if it is too awkward to avoid `fetchall()`, there may be little overall difference.

Regarding your observations:
1. Will not the cursor be automatically closed when it goes out of scope?
2. The lazy population of `index` is by design, as `Frame.from_records` must iterate over all records once before the `Index` is created. This pattern is used in a number of `Frame` constructors.

gdevanla: Thanks for explaining (2). 

On (1),  `close()` will not get called when cursor goes out of scope.  During GC, the cursor object itself could be dereferenced, but I think any clean needed between client and server (database), if implemented on GC is implementation dependent I think. I would this about a cursor as a file handle/external resource that is database dependent.

For example, I found that pysqlite has weakref's to the cursor objects and does a clean up once in a while. (this is based on my cursory understanding of what is happening in this library)
https://github.com/ghaering/pysqlite/blob/b945dafd4c3387c4e72dde2e0bf650c6a24cdc6e/src/connection.c#L774

And, I think for pysqlite, on dellocate up (reset is called on database).  (even if `close` is not called)
https://github.com/ghaering/pysqlite/blob/b945dafd4c3387c4e72dde2e0bf650c6a24cdc6e/src/cursor.c#L85

And, here is what is happening on `close`
https://github.com/ghaering/pysqlite/blob/b945dafd4c3387c4e72dde2e0bf650c6a24cdc6e/src/cursor.c#L904

But, I am not sure if it is guaranteed for, say psycopg2 or any other DBAPI compliant library.

flexatone: Great, thanks for these details. I am happy to merge this as is then. Thanks for the PR!

flexatone: I do not expect there to be a significant benefit to the more-lazy approach, as from-records has to call `list` on generators anyway. But if you want to try implementing it that way, we can take a look. I assume that `fetchall` is generally expected to be on all cursors, right?

flexatone (approved):

PR #312, opened 2021-01-30
Adds multi-process support for reading in multiple labels in StoreZipParquet

chaburkland:
I'm leaving this as a draft for now, because there's a lot more to do.

Also, I am holding off on writing unit tests until I get confidence the approach I'm taking aligns with SF code styles/standards.


static_frame/core/store.py
```diff
@@ -216,11 +232,20 @@ def __init__(self,
...
+                        config.write_chunksize != self._default.write_chunksize):
+                    raise ErrorInitStoreConfig(f'config {label} has write_max_workers/chunksize inconsistent with default; align values and/or pass a default StoreConfig.')
                 self._map[label] = config
 
     def __getitem__(self, key: tp.Optional[tp.Hashable]) -> StoreConfig:
         return self._map.get(key, self._default)
 
+    def is_empty(self) -> bool:
+        return len(self._map) == 0
+
```
flexatone, line 248: I do not think we need this, as we would generally just use the default config if no per-frame configs are provided.

static_frame/core/store_zip.py
```diff
@@ -173,22 +206,56 @@ def read_many(self,
...
+                label_encoded = config_map.default.label_encode(label)
                 src = BytesIO(zf.read(label_encoded + self._EXT_CONTAINED))
-                yield container_type.from_parquet(
-                        src,
+
+                src_and_name = src, label
+
+                if single_config:
+                    # No need to duplicate config info here
+                    frames_parameters.append(src_and_name)
```
flexatone, line 224: `c` in this context will be the default config, if no per-frame configs were defined, so we do not need this branch, I think.

static_frame/core/store_zip.py
```diff
@@ -164,6 +166,37 @@ class StoreZipParquet(_StoreZip):
...
 
+    @staticmethod
+    def _build_frame_explicit(
+            src_and_name: tp.Tuple[BytesIO, tp.Hashable],
+            *,
+            index_depth: int,
+            columns_depth: int,
+            columns_select: int,
+            dtypes: DtypesSpecifier,
+            consolidate_blocks: bool,
```
flexatone, line 177: Maybe just pass the StoreConfig object, rather than these parameters?

static_frame/core/store.py
```diff
@@ -218,6 +234,12 @@ def __init__(self,
                 if (config.label_encoder != self._default.label_encoder or
                         config.label_decoder != self._default.label_decoder):
                     raise ErrorInitStoreConfig(f'config {label} has encoder/decoder inconsistent with default; align values and/or pass a default StoreConfig.')
+                if (config.read_max_workers != self._default.read_max_workers or
+                        config.read_chunksize != self._default.read_chunksize):
+                    raise ErrorInitStoreConfig(f'config {label} has read_max_workers/chunksize inconsistent with default; align values and/or pass a default StoreConfig.')
+                if (config.write_max_workers != self._default.write_max_workers or
+                        config.write_chunksize != self._default.write_chunksize):
+                    raise ErrorInitStoreConfig(f'config {label} has write_max_workers/chunksize inconsistent with default; align values and/or pass a default StoreConfig.')
```
flexatone, line 242: I wonder if we can consolidate all of these default checks into  a loop, where we just iterate over the attrs we require should be the same?

static_frame/core/store_zip.py
```diff
@@ -34,39 +79,84 @@ def labels(self, *,
...
+
+                    hashable_config = HashableConfigParams( # pylint: disable=no-value-for-parameter
+                            index_depth=c.index_depth,
+                            index_name_depth_level=c.index_name_depth_level,
+                            columns_depth=c.columns_depth,
+                            columns_name_depth_level=c.columns_name_depth_level,
+                            columns_select=c.columns_select,
+                            dtypes=c.dtypes,
+                            consolidate_blocks=c.consolidate_blocks,
+                    )
```
flexatone, line 109: Maybe `StoreConfig` can have a method `to_store_config_he` or similar that returns a hashable config; that way we encapsulate that extraction.

static_frame/core/store_zip.py
```diff
@@ -3,22 +3,67 @@
...
 from static_frame.core.store import StoreConfigMapInitializer
 from static_frame.core.util import AnyCallable
 from static_frame.core.container_util import container_to_exporter_attr
+from static_frame.core.util import DtypesSpecifier
+
+
+FrameConstructor = tp.Callable[[tp.Any], Frame]
+
+
+class HashableConfigParams(tp.NamedTuple):
```
flexatone, line 23: Maybe we call this `StoreConfigHE`, following the naming we use elsewhere. Also, as mentioned below, `StoreConfig` can implement an exporter to create these.

static_frame/core/store_zip.py
```diff
@@ -3,22 +3,67 @@
...
+    index_depth: int
+    index_name_depth_level: tp.Optional[tp.Union[int, tp.Iterable[int]]]
+    columns_depth: int
+    columns_name_depth_level: tp.Optional[tp.Union[int, tp.Iterable[int]]]
+    columns_select: tp.Optional[tp.Iterable[str]]
+    dtypes: DtypesSpecifier
+    consolidate_blocks: bool
+
+
+class BuildFrameParams(tp.NamedTuple):
```
flexatone, line 33: Maybe a better name for this is `FrameInitDeferred`, as we use this as a payload to defer initialization/construction of the `Frame`.

static_frame/core/store_zip.py
```diff
@@ -3,22 +3,67 @@
...
 
     _EXT: tp.FrozenSet[str] = frozenset(('.zip',))
     _EXT_CONTAINED: str = ''
 
+    @classmethod
+    def _container_type_to_constructor(cls, container_type: tp.Type[Frame]) -> FrameConstructor:
+        raise NotImplementedError
+
+    @classmethod
+    def _build_frame_explicit(cls,
```
flexatone, line 50: I am not sure "explicit" is the right word here, but I cannot come up with anything better right now.

flexatone (changes requested):
This is an elegant implementation! Just a few suggestion for renaming and consolidation.

static_frame/core/store_zip.py
flexatone: Maybe `_frame_from_bytes`?

static_frame/core/store_zip.py
```diff
@@ -3,22 +3,67 @@
...
+            src: bytes,
+            name: tp.Hashable,
+            config: HashableConfigParams,
+            constructor: FrameConstructor,
+        ) -> Frame:
+        raise NotImplementedError
+
+    @classmethod
+    def _build_frame_from_params(cls, params: BuildFrameParams) -> Frame:
```
flexatone, line 59: And this would be `_frame_from_frame_init_deferred`, perhaps.

chaburkland: In my recent commit I updated this name

chaburkland:
I struggled a bit to come up with an elegant approach to `StoreConfigHE`. Defining it explicitly was *extremely* verbose, and so I opted for a more dynamic approach. I'm very open to (and expecting) feedback!

chaburkland:
Also, sorry for the partial commit - I pushed prematurely

static_frame/core/store.py
```diff
@@ -142,6 +168,13 @@ def label_decode(self, label: str) -> tp.Hashable:
             return self.label_decoder(label)
         return label
 
+    def to_store_config_he(self) -> 'StoreConfigHE':
+        '''
+        Return a ``StoreConfigHE`` version of this StoreConfig.
+        '''
+        return StoreConfigHE(**{attr: getattr(self, attr) for attr in self.__slots__})  #type: ignore [no-untyped-call]
```
flexatone, line 175: I think we should exclude the not hashable attrs here, rather than in the `StoreConfigHE`. We can define the not hashable in this class, too.

static_frame/core/store.py
```diff
@@ -152,6 +185,45 @@ def label_decode(self, label: str) -> tp.Hashable:
         ]
 
 
+class StoreConfigHE(StoreConfig):
+    '''
+    A hash/equals subclass of :obj:`StoreConfig`, permiting usage in a Python set, dictionary, or other contexts where a hashable container is needed.
+
+    To support hashability, label_encoder/label_decoder are ignored
+    '''
+    __slots__ = ('_hash',)
```
flexatone, line 194: Maybe it is easier of `StoreConfigHE` is not a subclass of StoreConfig, as it has a subset of its attrs. Actually, now that I think about it `StoreConfig` should be a subclass of `StoreConfigHE`, as it expands the attrs available. We can then override `__hash__` in `StoreConfig` to raise.

static_frame/core/store.py
```diff
@@ -215,9 +297,11 @@ def __init__(self,
                 if not isinstance(config, self._DEFAULT.__class__):
                     raise ErrorInitStoreConfig(
                         f'unspported class {config}, must be {self._DEFAULT.__class__}')
-                if (config.label_encoder != self._default.label_encoder or
-                        config.label_decoder != self._default.label_decoder):
-                    raise ErrorInitStoreConfig(f'config {label} has encoder/decoder inconsistent with default; align values and/or pass a default StoreConfig.')
+
+                for attr in self._ALIGN_WITH_DEFAULT_ATTRS:
+                    if getattr(config, attr) != getattr(self._default, attr):
+                        raise ErrorInitStoreConfig(f'config {label} has {attr} inconsistent with default; align values and/or pass a default StoreConfig.')
```
flexatone, line 303: This is so much better!

flexatone (changes requested):
Nice changes. Just an idea about `StoreConfig`, `StoreConfigHE`'s relationship that might tighten up the interfaces.

static_frame/core/store.py
chaburkland: I wondered about that, but I wasn't sure. That will definitely clean things up.

static_frame/core/store.py
```diff
@@ -152,6 +185,94 @@ def label_decode(self, label: str) -> tp.Hashable:
...
+                self.trim_nadir == other.trim_nadir and
+                self.include_index == other.include_index and
+                self.include_index_name == other.include_index_name and
+                self.include_columns == other.include_columns and
+                self.include_columns_name == other.include_columns_name and
+                self.merge_hierarchical_labels == other.merge_hierarchical_labels and
+                self.read_max_workers == other.read_max_workers and
+                self.read_chunksize == other.read_chunksize and
+                self.write_max_workers == other.write_max_workers and
+                self.write_chunksize == other.write_chunksize
```
flexatone, line 228: Can we just iterate through slots to do this?

static_frame/core/store.py
```diff
@@ -152,6 +185,94 @@ def label_decode(self, label: str) -> tp.Hashable:
...
+        if dtypes_specifier is None :
+            return dtypes_specifier
+        if isinstance(dtypes_specifier, dict):
+            return dtypes_specifier.items()
+        if isinstance(dtypes_specifier, list):
+            return tuple(dtypes_specifier)
+        return dtypes_specifier # type: ignore [return-value]
+
+    def __hash__(self) -> int:
+        if self._hash is None:
```
flexatone, line 251: This approach of caching the hash I took in SeriesHE, FrameHE for potential performance improvements, but if it simplifies the design, perhaps we do not need it here.

chaburkland: I think that it's good to keep, since we potentially have to call something like `.items()` on a dictionary, so making it happen only once lazily seems like the best approach.

static_frame/core/store.py
```diff
@@ -147,86 +126,24 @@ def __init__(self, *,
...
-                self.read_chunksize == other.read_chunksize and
-                self.write_max_workers == other.write_max_workers and
-                self.write_chunksize == other.write_chunksize
-        )
+
+        for attr in self.__slots__:
+            if attr.startswith('_'):
+                continue
+            if getattr(self, attr) != getattr(other, attr):
+                return False
```
flexatone, line 144: Much better!

static_frame/core/store.py
```diff
@@ -273,6 +190,121 @@ def __hash__(self) -> int:
...
+    '''
+    A read-only container of parameters used by :obj:`Store` subclasses for reading from and writing to multi-table storage formats.
+    '''
+    label_encoder: tp.Optional[tp.Callable[[tp.Hashable], str]]
+    label_decoder: tp.Optional[tp.Callable[[str], tp.Hashable]]
+
+    __slots__ = (
+            'label_encoder',
+            'label_decoder',
+            )
```
flexatone, line 203: Is this implicitly a union of the parent class? I generally re-define all the slots needed in derived class, but if this is the union, it is an efficient representation.

flexatone (approved):
Many thanks for this PR!

Please confirm you are done with your changes and I will merge it in.

static_frame/core/store.py
chaburkland: Yeah - I want all my parent's slots + these two new ones.

chaburkland: Maybe except `_hash`, although that won't matter since I explicitly raise on `__hash__()` ...

chaburkland:
Just pushed my final commit/changes.

Issue #250, opened 2020-09-22
Re-thinking from_delimited


ForeverWintr:
In response to your call for api breaking suggestions, I'd like to suggest a change to the way SF reads delimited text. Namely, I think we should split the *reading* from the type guessing that happens in genfromtxt. 

In many places in SF, we embrace the zen of python's "In the face of ambiguity, refuse the temptation to guess", but the `from_delimited` family of constructors does not. Instead it follows pandas' lead by attempting to guess the type of the data it reads. As we've seen in both pandas and static frame (#130, #160), this causes issues when the guessed type is unexpected.  

We tend to think of text files as 'untyped', but of course they're actually typed as strings! Currently there's no way in static frame to simply read a frame from a text file without performing type guessing. I think we should change this, and not perform type guessing unless explicitly instructed to do so. 

I can think of a couple of ways to implement this. One is a new method that takes a frame and returns a dtype guess for each column (e.g., as a mapping from column name to dtype):

```python
strings = sf.Frame.from_tsv('/tmp/tmp.tsv')
guessed_dtypes = strings.guess_dtypes()
typed = sf.Frame(strings, dtypes=guessed_dtypes)
```

Another is a new argument to the `from_delimited` constructor that explicitly tells sf which columns to guess types for:

```python
# the following guesses the dtype of 'column_a' and 'column_c' by inspecting their contents.
# It does not inspect 'column_b', but just converts it to int.
typed = sf.Frame(strings, guess_types=['column_a', 'column_c'], dtypes={'column_b': int})
```

The second may be more efficient (as it can avoid the intermediate Frame of strings), the first has other benefits, like giving users a way to reinterpret types that were coerced to `object`.

# Speed benefits

A secondary benefit to separating type guessing from data loading is speed. Static frame's type guessing is the largest contributor to its slow text loading when compared to pandas. In [my profiling](https://github.com/InvestmentSystems/static-frame/files/5258913/profile_results.txt), it accounts for 98.8% of total runtime! I wrote a *simple as possible* text reader (<20 lines) that simply reads the contents of a delimited io object into a numpy array, and uses that to create a frame. You can see its implementation (and my performance comparison code) [on this branch](https://github.com/ForeverWintr/static-frame/tree/from_delimited). 

[...]

PR #322, opened 2021-03-01
From delimited

ForeverWintr:
This is a WIP implementation of a `from_delimited_no_guess`, which functions similarly to `from_delimited` but without type guessing. It foregoes `np.genfromtxt`, and instead uses the `csv` module exclusively. Once strings have been parsed, it avoids modifying them unless instructed to do so with `dtypes` and `converters` arguments. 

The core functionality is complete, but the tests are failing because I temporarily re-targeted the regular `from_delimited` to use my approach. I was going through the existing tests and making  `from_delimited_no_guess` versions of many of them.

refs #250


flexatone:
Many thanks for this PR! The approach taken here is similar to what I am working on as a C extension in ArrayKit, so I am happy to see this and will continue to study it. The specialized performance tests will also be useful.

flexatone:
Any thoughts on how you would have handled a case like this:
```python
>>> np.genfromtxt(['False|True|3.4|(3+2j)'], delimiter='|', dtype=None)                                                            
array((False, True, 3.4, 3.+2.j),
      dtype=[('f0', '?'), ('f1', '?'), ('f2', '<f8'), ('f3', '<c16')])
```

If we have only loaded strings, we cannot just `astype` to object:

```python
>>> np.array(('False', 'True', '3.4' , '(3+2j)')).astype(object)                                                                   
array(['False', 'True', '3.4', '(3+2j)'], dtype=object)
```
So assume you intended to handle this sort of case with a `converters` function?

ForeverWintr:
Yes, cases like that are what made me realize a `converters` function was worthwhile. In this case you'd use it like this (writing from memory)

```python
def istrue(s):
    return s == 'True'

sf.Frame.from_delimited_no_guess(
    fp,
    delimiter='|',
    dtypes={'a': np.bool_, "b": np.bool_, 'c': np.float64, 'd': np.complex,},
    converters={'a': istrue, 'b': istrue,'d': complex,},
)
```

flexatone:
Thanks for that example.

But I was imagining that those values were all in one column: the converters function would be rather complex in that case, no?

ForeverWintr:
I see! The converter would indeed be complex in that case. Are you aware of a better option?

In general, I think we should optimize interfaces for the common case, and that case seems uncommon. I added converters because I think there are some common, simple conversions users will need. For more advanced transformations, I think it's fine to deliver the user a frame of strings and let them apply their own conversion code. The fact that they can rely on getting strings for any column they haven't explicitly converted makes that easier.

ForeverWintr:
I'll close this, as I believe it is superseded by #651

PR #335, opened 2021-04-14
Start shipping a `py.typed` file.

brandtbucher:
Currently, all symbols imported from `static_frame` in user code are seen by type-checkers as `typing.Any`.

The presence of this file in a distribution indicates to tools like mypy that the annotations in `static_frame` can actually be used when linting user code.

https://mypy.readthedocs.io/en/stable/installed_packages.html#making-pep-561-compatible-packages


flexatone (approved):

flexatone:
Thanks for this addition!

PR #336, opened 2021-04-18
integration of AK 0.1.2

flexatone:


static_frame/core/util.py
```diff
@@ -288,66 +288,80 @@ class PairRight(Pair):
...
-
-def row_1d_filter(array: np.ndarray) -> np.ndarray:
-    '''
-    Ensure that a row that might be 2D or 1D is returned as a 1D array.
-    '''
-    if array.ndim == 2:
-        # could assert that array.shape[0] == 1, but this will raise if does not fit
-        return np.reshape(array, array.shape[1])
-    return array
+from arraykit import mloc
```
brandtbucher, line 291: FYI, this is how you tell mypy to "reexport" an import:

```suggestion
from arraykit import mloc as mloc
```

Just adding "as" with the same name should do it.

I also just realized that we should probably add a `py.typed` file to ArrayKit, just as we did here.

flexatone: Thanks, that does the trick!

PR #339, opened 2021-05-18
Allow unsetting index when columns are IH

ForeverWintr:
Currently we raise `ErrorInitFrame('cannot unset index with a columns with depth greater than 1')` if you try to `unset_index` on a frame with hierarchical columns. I run into this occasionally, and would like to propose this change, which allows you to unset index with a column hierarchy so long as you provide names with the right number of levels.

What do you think?


static_frame/core/frame.py
```diff
@@ -5054,14 +5054,14 @@ def blocks() -> tp.Iterator[np.ndarray]:
...
-            raise ErrorInitFrame('cannot unset index with a columns with depth greater than 1')
-
         if names:
             columns = chain(names, self._columns.values)
         else:
             columns = chain(self._index.names, self._columns.values)
 
+        if self._columns.depth > 1:
+            columns = IndexHierarchy.from_labels(columns)
+
```
flexatone, line 5064: Here, we need to be pass on the same composed index constructors that are already used in the `Index`.

```suggestion
        if self._columns.depth > 1:
            index_constructors = tuple(self._index._levels.index_types())
            columns = IndexHierarchy.from_labels(columns, index_constructors=index_constructors)

```

Also, this PR suggests a different approach to ``blocks()``, above. This permits us to use the underlying arrays from the `IndexHierarchy` without doing type consolidation into a single array.

```python
        def blocks() -> tp.Iterator[np.ndarray]:
            if self._index.ndim == 1:
                yield self._index.values
            else:
                if self._index._recache:
                    self._index._update_array_cache()
                yield from self._index._blocks._blocks
            for b in self._blocks._blocks:
                yield b
```

Note that SF commonly uses private interfaces of component objects in functionality like this to avoid creating unnecessary intermediary data structures.

flexatone (changes requested):
Many thanks for this PR! This is a good enhancement. My comments suggest some changes to address reuse and retaining composed `Index` types. Please add a test with an `IndexHierarchy` with a composed `IndexDate` that shows that the `IndexDate` is preserved after calling `unset_index()`.

ForeverWintr:
Thanks for the review @flexatone! Can you elaborate on what you mean by a 'composed' IndexDate?

Also, when I apply your proposed change to the nested blocks function, `test_frame_pivot_f` fails because the post-pivot dtypes are not as the test expects. I think the new version may be correct, as the resulting dtypes appear to fit the frame's values better, but perhaps you can take a look? I'll commit a passing version of that test so you can see how the result changes.

flexatone:
@ForeverWintr : An `IndexHierarchy` has an `Index` type per depth level, such that different `Index` types are "composed" inside the `IndexHierarchy`. So I just mean create an `IndexHierarchy` that has an `IndexDate` as one of its depth index types.

I will take a look at this test; thanks!

flexatone:
Please also update ``new.rst`` and document this change for the 0.8.13.

ForeverWintr:
@flexatone I've added a test with a composed IH containing dates, but unfortunately dtypes are not preserved. They're lost in the call to `self._columns.values`, which combines everything into a single array. Perhaps the solution is to use the underlying TypeBlocks directly?

flexatone:
@ForeverWintr : yes, if you can avoid using `self._columns.values` and operate on `self._columns._blocks` there hsould be a way to preserve the dtypes.

ForeverWintr:
@flexatone I'm able to get a collection of numpy arrays with dtype preserved from the `_columns._blocks._blocks`, but I don't think there's an existing `IndexHierarchy` constructor that consumes them in this form. `from_labels` interprets each array as a row, but I need each array as a column. I don't see an existing method to construct an IndexHierarchy directly from blocks.

Could you help me understand why the `IndexHierarchy` constructor needs an `IndexLevel` instance, and how I can construct one from `blocks`?

Also, I think a getter for the `_blocks` attribute that recaches if needed would be a worthwhile addition. It took me a while to figure out why `self._columns._blocks._blocks` raised an `AttributeError` unless I printed `self._columns` first.

static_frame/core/frame.py
```diff
@@ -5097,13 +5103,19 @@ def blocks() -> tp.Iterator[np.ndarray]:
         else:
             block_gen = blocks
 
-        if self._columns.depth > 1:
-            raise ErrorInitFrame('cannot unset index with a columns with depth greater than 1')
+        # self._columns._blocks may be None until array cache is updated.
+        self._columns._update_array_cache()
+        column_blocks = self._columns._blocks._blocks
```
flexatone, line 5108: Here I would follow pattern as above and check the recache status first:

```suggestion
        if self._columns._recache:
            self._columns._update_array_cache()
        column_blocks = self._columns._blocks._blocks
```

static_frame/core/frame.py
```diff
@@ -5097,13 +5103,19 @@ def blocks() -> tp.Iterator[np.ndarray]:
...
 
         if names:
-            columns = chain(names, self._columns.values)
+            columns = chain(np.array(tuple(names)), column_blocks)
         else:
-            columns = chain(self._index.names, self._columns.values)
+            columns = chain(self._index.names, column_blocks)
```
flexatone, line 5113: I am not sure this is right.

The blocks obtained from the `IndexHierarchy` are along the axis of the types, which for an `IndexHierarchy` in the columns positions are oriented like "rows":

```python
>>> ih = sf.IndexHierarchy.from_product(('a', 'b'), (1, 2))                                                                                                                                              
>>> ih                                                                                                                                                                                                   
<IndexHierarchy>
a                1
a                2
b                1
b                2
<<U1>            <int64>
>>> ih.values # force a recache call                                                                                                                                                                     
array([['a', 1],
       ['a', 2],
       ['b', 1],
       ['b', 2]], dtype=object)
>>> ih._blocks._blocks                                                                                                                                                                                   
[array(['a', 'a', 'b', 'b'], dtype='<U1'), array([1, 2, 1, 2])]
```
So if the user is providing `names`, and the number of `names` matches the depth of the old index, than we will want to do something like this to concatenate the names in front of the blocks. But here I am duplicating the names for each level of the columns, which is a little strange but I am not sure what else to do.

```python
column_blocks = [concat_resolved(np.array(names), block) for block in blocks]
```
Then you can create the columns with:
```python
columns = self._COLUMNS_HIERARCHY_CONSTRUCTOR._from_type_blocks(TypeBlocks.from_blocks(column_blocks), own_blocks=True)
```

ForeverWintr: Thanks! I was fairly confident that what I was doing wasn't correct. I've updated the code based on your suggestion, and am able to correctly preserve dtypes. 

I had to adapt your suggestion, as `concat_resolved` expects 2d arrays. Here's the result:

```python
        column_blocks = self._columns._blocks._blocks
        names_t = zip(*names)

        column_blocks_new = tuple(
                concat_resolved((np.array([name]), block[np.newaxis]), axis=1).T
                for name, block in zip(names_t, column_blocks)
        )
        column_type_blocks = TypeBlocks.from_blocks(column_blocks_new)
```

ForeverWintr:
The next issue I'm running into is that `IndexGO` has no `_blocks`. Any suggestions on how to handle this?

flexatone:
@ForeverWintr : generally we branch based on the `.depth` attr; if `depth == 1`, we know we have a 1D structure and can simply treat its `.values` attr as a 1D block.

ForeverWintr:
@flexatone Sorry for the delay; I believe this is functioning correctly now. 

I see some quality and forward test failures, but those I can reproduce are happening on master too, so I don't think they're related to my changes. Can you take a look?

flexatone (approved):
Very nice  work on this. Thank you for this contribution.

PR #368, opened 2021-08-26
Quilt Repr & Error Handling Improvements

chaburkland:
This is a PR to address QoL improvements for Quilt creation errors, as well as providing a frame-like repr to the class.

The main error improvements are as follows:

1. `from_frames`, where frames do not have unique names.

Current behavior
```python
>>> import frame_fixtures as ff

>>> f1 = ff.parse("s(3,3)")
>>> f2 = ff.parse("s(3,3)")

>>> sf.Quilt.from_frames((f1, f2), ...)
ErrorInitIndexNonUnique: labels (2) have non-unique values (1)
```
New behavior:
```python
ErrorInitIndexNonUnique: Frames do not have unique names
```

2. Quilts will raise more descriptive failures when frames do not align on `axis` based on `retain_labels`

Current behavior
```python
# Misalignment on index or columns
ErrorInitIndexNonUnique: labels (n) have non-unique values (m)
```
New behavior:
```python
# Misalignment on index
ErrorInitQuilt: Duplicate index labels across frames. Either ensure all indices are unique for all frames, or set retain_labels=True to obtain an IndexHierarchy

# Misalignment on columns
ErrorInitQuilt: Duplicate column labels across frames. Either ensure all columns are unique for all frames, or set retain_labels=True to obtain an IndexHierarchy
```
--------------------------

The biggest change by far, is the change to how the Quilt's repr is displayed.
Previously, the repr was like this:
```python
<Quilt at 0x...>
or 
<Quilt: name at 0x...>
```

Now, it looks like so
1. Quilt aligned on `axis=0`, where `retain_labels=True`
```python
<Quilt>
<Index: Aligned Columns>     A   B   ... Y   Z
<Index: Concatenated Frames>         ...             
frame_1                      .   .   ... .   .
frame_2                      .   .   ... .   .
...                          ... ... ... ... ...
frame_n-1                    .   .   ... .   .
frame_n                      .   .   ... .   .
```

2. Quilt aligned on `axis=1`, where `retain_labels=True`
```python
<Quilt>
<Index: Concatenated Frames> frame_1   frame_2   ...  frame_n-1 frame_n
<Index: Aligned Indices>                         ...                       
0                            .         .         ...  .         .
1                            .         .         ...  .         .
...                          ...       ...       ...  ...       ...
98                           .         .         ...  .         .
99                           .         .         ...  .         .
```

3. Quilt aligned on `axis=0`, where `retain_labels=False`
```python
<Quilt>
<Index: Aligned Columns>      A   B   ... Y   Z
<Index: Concatenated Indices>         ...       
0                             .   .   ... .   .
1                             .   .   ... .   .
...                           ... ... ... ... ...
198                           .   .   ... .   .
199                           .   .   ... .   .
```

4. Quilt aligned on `axis=1`, where `retain_labels=False`
```python
<Quilt>
<Index: Concatenated Columns> A   B   ... Y   Z
<Index: Aligned Indices>              ...         
0                             .   .   ... .   .
1                             .   .   ... .   .
...                           ... ... ... ... ...
98                            .   .   ... .   .
99                            .   .   ... .   .
```

Final notes: I haven't fully tested how this new approach interacts with frames that already have`IndexHierarchy`'s.


static_frame/core/axis_map.py
```diff
@@ -37,31 +38,38 @@ def bus_to_hierarchy(
...
+                    # What about same labels different order?
                     raise init_exception_cls('opposite axis must have equivalent indices')
         elif axis == 1:
-            tree[label] = extractor(f.columns)
+            tree[label] = tree_extractor(f.columns)
             if opposite is None:
                 opposite = extractor(f.index)
             else:
                 if not opposite.equals(f.index):
+                    # What about same labels different order?
```
flexatone, line 67: At this point we are not reindexing in any form (even just for ordering) on the opposite axis, but we might in the future.

static_frame/core/bus.py
```diff
@@ -97,11 +98,15 @@ def from_frames(cls,
...
-                    name=name,
-                    )
+        try:
+            series = Series.from_items(
+                        ((f.name, f) for f in frames),
+                        dtype=DTYPE_OBJECT,
+                        name=name,
+                        )
+        except ErrorInitIndexNonUnique:
+            raise ErrorInitIndexNonUnique("Frames do not have unique names") from None
```
flexatone, line 108: Nice Exception improvement.

static_frame/core/display.py
```diff
@@ -22,6 +23,9 @@
 from static_frame.core.display_config import _DISPLAY_FORMAT_TERMINAL
 from static_frame.core.style_config import StyleConfig
 
+if tp.TYPE_CHECKING:
+    from static_frame.core.index_base import IndexBase # pylint: disable=unused-import
```
flexatone, line 27: Please add coverage exclusion too. 
```suggestion
    from static_frame.core.index_base import IndexBase # pylint: disable=unused-import #pragma: no cover
```

static_frame/core/index_hierarchy.py
```diff
@@ -1565,6 +1568,37 @@ def to_pandas(self) -> 'DataFrame':
...
+        for labels in self.iter_label():
+            add_labels(tree, labels)
+
+        def clean(tree: TreeNodeGrowableT) -> None:
+            # Make the leafs immutable!
+            for k in tuple(tree.keys()):
+                if isinstance(tree[k], list):
+                    tree[k] = tuple(tree[k])
+                else:
+                    clean(tree[k])
```
flexatone, line 1597: As the tree is just a temporary container between index objects, do we need to do a full-depth transversal to convert `list` to `tuple`? Or, can it be done in the initial traversal?

static_frame/core/index_hierarchy.py
```diff
@@ -1638,7 +1674,7 @@ def level_drop(self,
                 blocks = self._blocks.iloc[NULL_SLICE, :count]
                 return self.__class__(levels,
                         blocks=blocks,
-                        own_blocks=True,
+                        own_blocks=True, # Should this be false?
```
flexatone, line 1677: We can own the blocks here as a the `self._blocks.iloc` call above gave us a new `TypeBlocks` instance.

static_frame/core/index_hierarchy.py
```diff
@@ -1665,7 +1701,7 @@ def level_drop(self,
                 return self.__class__(levels,
                         name=name,
                         blocks=blocks,
-                        own_blocks=True
+                        own_blocks=True, # Should this be false?
```
flexatone, line 1704: Same as above.

static_frame/core/quilt.py
```diff
@@ -426,15 +426,25 @@ def _update_axis_labels(self) -> None:
                     init_exception_cls=ErrorInitQuilt,
                     )
 
+        axis_label = 'index' if self._axis == 0 else 'column'
+        axis_labels = 'indices' if self._axis == 0 else 'columns'
+        err_msg = f"Duplicate {axis_label} labels across frames. Either ensure all {axis_labels} are unique for all frames, or set retain_labels=True to obtain an IndexHierarchy"
```
flexatone, line 431: I like the improved `Exception` message this provides, but maybe this should be tucked away in a function only called when the `Exception` is raised. Always doing it for non-exceptional cases seems unnecessary.

static_frame/core/quilt.py
```diff
@@ -466,29 +476,67 @@ def rename(self, name: NameType) -> 'Quilt':
...
+
+        Args:
+            {config}
         '''
         if self._assign_axis:
             self._update_axis_labels()
-        return self.to_frame().display( #type: ignore
-                config,
+
+        bus_columns = False # Flag to strip off column dtype info
```
flexatone, line 493: Maybe this can have a more explicit name, such as `drop_first_row_last_item` or similar.

static_frame/core/quilt.py
```diff
@@ -466,29 +476,67 @@ def rename(self, name: NameType) -> 'Quilt':
...
+            if bus_columns:
+                # First Column Row -> last element is the dtype of the column
+                # Guaranteed to not be index hierarchy as buses cannot have index hierarchies
+                d._rows[1].pop()
+
+            # Since placeholder_gen is not a ndarray, there is no dtype to append in the final row
+            # However, in the case of a center ellipsis being added, an ellipsis will be
+            # awkwardly placed direclty adjacent to the index dtype information.
+            if d._rows[-1][-1] is Display.CELL_ELLIPSIS:
+                d._rows[-1].pop()
```
flexatone, line 537: I wish there was a better way, but I understand there may not be at this time.

I think we want to use string equality here rather than object id.

```suggestion
            if d._rows[-1][-1] == Display.CELL_ELLIPSIS:
                d._rows[-1].pop()
```

flexatone (changes requested):
Very nice work with these valuable improvements. Just a few small changes requested. The new tests look great, but please check the coverage report to ensure we are not loosing coverage any any of the affected modules.

static_frame/core/display.py
chaburkland: This change came in from default :+1:

static_frame/core/index_hierarchy.py
chaburkland: Probably -> but that would require me to know how `IndexLevels` work...

The current algorithm iterates through each label of the index hierarchy, so I don't have access to the innermost index objects.

Let me take a look to see if I how to iterate more intelligently.

PR #397, opened 2021-11-08
Adds relabel_at_depth to IndexHierarchy.

### LINKED ISSUE #56: IndexHierarchy.relabel_at_depth
Permit modifying an IndexHierarchy at a specified, integer depth.

chaburkland:
Closes #56


static_frame/core/index_hierarchy.py
```diff
@@ -1051,14 +1050,65 @@ def gen() -> tp.Iterator[tp.Tuple[tp.Hashable, ...]]:
...
+
+        if len(depth_level) == 1:
+            # Optimized path for single depth level
+            return self.__class__(
+                levels=self._levels.relabel_at_specific_depth(mapper, depth_level[0]),
+                name=self._name,
+                )
+        elif len(depth_level) == self.depth:
+            # Equivalent to relabel!
+            return self.relabel(mapper)
```
flexatone, line 1080: What if the `depth_level` is invalid (i.e., `[1, 1, 1]`) or in a different order than expected?

static_frame/core/index_level.py
```diff
@@ -372,6 +372,37 @@ def get_labels(index: Index,
                 next_depth = depth + 1
                 levels.extend((lvl, next_depth) for lvl in level.targets)
 
+    def relabel_at_specific_depth(self,
+            mapper: RelabelInput, depth_level: int
+            ) -> 'IndexLevel':
```
flexatone, line 377: Maybe rename to "single" depth?

```suggestion
    def relabel_at_single_depth(self,
            mapper: RelabelInput, 
            depth_level: int,
            ) -> 'IndexLevel':
```

static_frame/core/index_level.py
```diff
@@ -372,6 +372,37 @@ def get_labels(index: Index,
...
+                targets=self.targets,
+                offset=self.offset,
+                own_index=True,
+                depth_reference=self._depth
+            )
+
+        if self.targets is None:
+            raise ValueError(f'cannot relabel at depth {depth_level}')
+
+        new_targets = np.empty(len(self.targets), dtype=object)
```
flexatone, line 395: We have a constant for this:
```suggestion
        new_targets = np.empty(len(self.targets), dtype=DTYPE_OBJECT)
```

static_frame/core/index_level.py
```diff
@@ -372,6 +372,37 @@ def get_labels(index: Index,
...
+        for i, target in enumerate(self.targets):
+            new_targets[i] =  target.relabel_at_specific_depth(mapper, depth_level - 1)
```
flexatone, line 397: This recursive implementation is startling simple. Nice work!

static_frame/core/index_hierarchy.py
```diff
@@ -1051,14 +1050,65 @@ def gen() -> tp.Iterator[tp.Tuple[tp.Hashable, ...]]:
...
+
+        index_constructors = tuple(self._levels.index_types())
+
+        def get_modified(array_items: tuple) -> tp.Iterable[tp.Any]: # type: ignore
+            if callable(mapper):
+                return mapper(array_items) # type: ignore
+
+            # if a mapper, it must support both __getitem__ and __contains__
+            return mapper.get(array_items, array_items) # type: ignore
```
flexatone, line 1089: I think our expectation is that, if multiple depth levels are provided, the same mapper will be applied on each element in each depth. It looks here that we are mapping/processing the entire row, which is plausible interpretation of what this function does, but maybe not something we need to support.

static_frame/core/index_hierarchy.py
```diff
@@ -1051,14 +1050,65 @@ def gen() -> tp.Iterator[tp.Tuple[tp.Hashable, ...]]:
...
+
+        array_getter = itemgetter(*depth_level)
+
+        def gen() -> tp.Iterator[np.ndarray]:
+            for array in self._blocks.axis_values(axis=1):
```
flexatone, line 1094: This will force type consolidation across the columns of the IH, which can lead to some undesirable coercions. If we make this function only operate on elements of each depth level in isolation, we can avoid this and perhaps use / extend your approach on `self._levels`.

flexatone (changes requested):
Many thanks for working on this! I think we can simplify what this function does if we approach the mapper/function application as always processing single labels, even if multiple depth-levels are specified: we just apply the same mapper/func to elements in the specified depths.

Also, be sure to test if this works as expected for `IndexHierarchyGO`.

static_frame/core/index_hierarchy.py
```diff
@@ -1020,7 +1021,7 @@ def index_types(self) -> 'Series':
         return Series(self._levels.index_types(), index=labels, dtype=DTYPE_OBJECT)
 
     #---------------------------------------------------------------------------
-    def relabel(self, mapper: RelabelInput) -> 'IndexHierarchy':
+    def relabel(self, mapper: RelabelInput) -> "IndexHierarchy":
```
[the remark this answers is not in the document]
chaburkland: Haha. I didn't mean to do this

static_frame/core/index_hierarchy.py
```diff
@@ -1031,13 +1032,11 @@ def relabel(self, mapper: RelabelInput) -> 'IndexHierarchy':
...
         if not callable(mapper):
             # if a mapper, it must support both __getitem__ and __contains__
-            getitem = getattr(mapper, 'get')
-
             def gen() -> tp.Iterator[tp.Tuple[tp.Hashable, ...]]:
                 for array in self._blocks.axis_values(axis=1):
                     # as np.ndarray are not hashable, must tuplize
                     label = tuple(array)
-                    yield getitem(label, label)
+                    yield mapper.get(label, label) # type: ignore
```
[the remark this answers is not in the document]
chaburkland: It was to avoid having a `getattr` call, and not being convinced when I was reading it that the performance advantage would be noticeable. I can put it back how it was though

static_frame/core/index_hierarchy.py
chaburkland: Ah I didn't consider that edge case. Good suggestion

static_frame/core/index_level.py
chaburkland: Thanks. I'm not sure what it would mean to apply a `mapper` to multiple depths?

static_frame/core/index_hierarchy.py
chaburkland: Ah okay that's what @nekobon said she expected too. Although my interpretation of the commented out stub was row-based. If that's incorrect, then I will update it. Are you wanting the other implementation? I'm not exactly sure what that would mean though. For example, how would you specify a relabel at multiple levels using a dictionary? Would you provide a sequence of dicts?

```python
ih.relabel_at_depth([dict(A=1), dict(I="i")), levels=[0,1])
```
?

static_frame/core/index_hierarchy.py
chaburkland: This is the exact same approach used in `IndexHierarchy.relabel`. I opted for this as I could not think of an alternative way to apply a mapping that's row-based on a subset of the index depths.

chaburkland:
@flexatone Based on the comments, I think it's worth meeting to discuss this issue more, as I think I have a fundamental misunderstanding of what is means to relabel at multiple depths

static_frame/core/index_hierarchy.py
```diff
@@ -1051,14 +1052,70 @@ def gen() -> tp.Iterator[tp.Tuple[tp.Hashable, ...]]:
...
+            mapper: RelabelInput,
+            depth_level: DepthLevelSpecifier = 0
+            ) -> "IndexHierarchy":
+        '''
+        Return a new :obj:`IndexHierarchy` after applying `mapper` to a level or each individual level specified by `depth_level`.
+
+        `mapper` can be a callable, mapping, or iterable.
+            - If a callable, it must accept a single value, and return a single value.
+            - If a mapping, it must map a single value to a single value.
+            - If a iterable, it must be the same length as `self`.
```
flexatone, line 1065: I am not sure if this list presentation translates well in the RST docs... do you know? If not, we can just make it a sentence.

static_frame/core/index_hierarchy.py
```diff
@@ -1051,14 +1052,70 @@ def gen() -> tp.Iterator[tp.Tuple[tp.Hashable, ...]]:
...
+        if callable(mapper) or hasattr(mapper, 'get'):
+            return self.__class__(
+                    levels=self._levels.relabel_at_depth(mapper, depth_level),
+                    name=self._name,
+                    )
+
+        values, _ = iterable_to_array_1d(mapper, count=len(self))
+
+        if len(values) != len(self):
+            raise ValueError('mapper must return a value for each label')
```
flexatone, line 1104:
```suggestion
            raise ValueError('Iterable must provide a value for each label')
```

static_frame/core/index_hierarchy.py
```diff
@@ -1051,14 +1052,70 @@ def gen() -> tp.Iterator[tp.Tuple[tp.Hashable, ...]]:
...
+            for depth_idx in range(self.depth):
+                if depth_idx in target_depths:
+                    yield values
+                else:
+                    yield self._blocks._extract_array_column(depth_idx)
+
+        return self.__class__._from_type_blocks(
+                TypeBlocks.from_blocks(gen()),
+                name=self._name,
+                index_constructors=tuple(self._levels.index_types()),
```
flexatone, line 1116: This suggests that this function in the future will need to take an `index_constructors` argument: a caller, for example, might provide values as a `np.datetime64` array, and thus need to provide `index_constructors` for each `depth_level` given. This can be deferred if you want to just get in what we have now.

static_frame/core/index_level.py
```diff
@@ -372,6 +371,64 @@ def get_labels(index: Index,
...
+
+        new_targets = np.empty(len(self.targets), dtype=DTYPE_OBJECT)
+        for i, target in enumerate(self.targets):
+            new_targets[i] = target.relabel_at_depth(mapper, target_depths)
+
+        return self.__class__(
+            index=index,
+            targets=ArrayGO(new_targets, own_iterable=True),
+            offset=self.offset,
+            own_index=False,
```
flexatone, line 429: You can `own_index` at least for the scenario on line 415; you can assign `own_index` on that branch, perhaps.

static_frame/test/unit/test_index_hierarchy.py
```diff
@@ -1355,7 +1308,218 @@ def test_hierarchy_relabel_a(self) -> None:
...
+
+        # Depth level outside range negative
+        with self.assertRaises(ValueError):
+            ih.relabel_at_depth(range(4), depth_level=-1)
+
+        # No depth levels!
+        with self.assertRaises(ValueError):
+            ih.relabel_at_depth(lambda:None, depth_level=[])
+
+    def test_hierarchy_relabel_at_depth_properties(self) -> None:
```
flexatone, line 1481: This is a great test. Maybe just make another version of the same with an `IndexHierarchyGO` to compare these assumptions in that context.

flexatone (changes requested):
This more narrow functionality is nice improvement. Please see my notes on taking an `index_constructors` arg and `IndexHierarchyGO` testing.

static_frame/core/index_level.py
chaburkland: Ah good catch

static_frame/core/index_hierarchy.py
chaburkland: Good idea! Let me add this as an optional argument to this method and `relabel`

chaburkland: Actually on second that, this is a bit more involved than I thought, so I will leave as is, but create a backlog issue

static_frame/core/index_hierarchy.py
chaburkland: I just did a fresh doc build, and it renders nicely, so I will leave it as is :+1:

flexatone (approved):
Nice work, and thanks for all the additional cleanup.

PR #437, opened 2022-02-09
Burkland/index hierarchy overhaul

chaburkland:
Massive overhaul to how IndexHierarchies are designed.

There are some side effects this change will bring about, that I will list below as I iterate more on this branch.

This is still very priliminary, as I do not have strong evidence yet that this new approach makes static-frame faster.


flexatone:
Many thanks for this PR. I look forward to seeing how it evolves.

Why so you have so many white-space removals in test module not related to this PR? It makes it look the changes here are much broader than they are.

chaburkland:
> `IndexHierarchyOld`



> Why do you have so many white-space removals in test module not related to this PR? It makes it look the changes here are much broader than they are.

Yeah I realized that this morning that this really clutters the diff. I'm making a separate branch right now that will isolate out the whitespace only changes. I just got carried away in my quest for uniformity

static_frame/core/index_hierarchy.py
```diff
@@ -953,19 +1001,19 @@ def _ufunc_set(self,
                 index_constructors=index_constructors,
                 depth_reference=self.depth)
 
-
     #---------------------------------------------------------------------------
+
+    @property
+    def _index_constructors(self) -> tp.Tuple[tp.Type[Index], ...]:
+        return tuple(index.__class__ for index in self._indices)
```
flexatone, line 1008: You might have this be an iterator of `tp.Type[Index]` so that the caller can decide if they need a tuple or not.

static_frame/core/index_hierarchy.py
```diff
@@ -456,129 +558,129 @@ def _from_type_blocks(cls: tp.Type[IH],
...
     def __init__(self,
-            levels: tp.Union[IndexLevel, 'IndexHierarchy'],
+            indices: tp.Union["IndexHierarchy", tp.List[Index]],
             *,
+            indexers: tp.List[np.ndarray] = (), # type: ignore
             name: NameType = NAME_DEFAULT,
-            blocks: tp.Optional[TypeBlocks] = None,
-            own_blocks: bool = False,
+            _blocks: tp.Optional[TypeBlocks] = None,
+            _own_blocks: bool = False,
```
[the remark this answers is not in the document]
chaburkland: Okay - there's enough checks in `__init__` now to ensure it is instantiated correctly, so I can remove the leading underscore.

chaburkland:
IndexHierarchy - Changes

# Behavior Changes
# **sf.Yarn**
### Change 1
#### Setup
```python
f1 = ff.parse('s(4,4)|v(int,float)').rename('f1')
f2 = ff.parse('s(4,4)|v(str)').rename('f2')
f3 = ff.parse('s(4,4)|v(bool)').rename('f3')
b1 = sf.Bus.from_frames((f1, f2, f3), name='a')

f4 = ff.parse('s(4,4)|v(int,float)').rename('f4')
f5 = ff.parse('s(4,4)|v(str)').rename('f5')
b2 = sf.Bus.from_frames((f4, f5), name='b')

f6 = ff.parse('s(2,4)|v(int,float)').rename('f6')
f7 = ff.parse('s(4,2)|v(str)').rename('f7')
b3 = sf.Bus.from_frames((f6, f7), name='c')

yarn = sf.Yarn.from_buses((b1, b2, b3), retain_labels=False, name='foo')
```
#### Old Behavior:
```python
with self.assertRaises(ErrorInitIndex):
    y1.loc[['f1', 'f7', 'f3']]
```
#### New Behavior:
```python
print(yarn.loc[['f1', 'f7', 'f3']].shapes)
<Series: shape>
<Index>
f1              (4, 4)
f7              (4, 4)
f3              (4, 2)
<<U2>           <object>
```
#### Rationale
Not limited to indices that follow a strict tree hierarchy

# **sf.Frame**
### Change 1
#### Setup
```python
f1 = sf.Frame.from_dict_records(
        [dict(a=1,b=1), dict(a=2,b=3), dict(a=1,b=1), dict(a=2,b=3)],
        index=sf.IndexHierarchy.from_labels(
            [(1,'dd',0),(1,'b',0),(2,'cc',0),(2,'ee',0)]
        )
)

series = f1.loc[sf.HLoc[(1,'dd')]]
frame = f1.loc[sf.HLoc[(1,['dd'])]]
print(series)
print(frame)
```
### Old Behavior:
```python
<Frame>
<Index>                        a       b       <<U1>
<IndexHierarchy>
1                dd    0       1       1
<int64>          <<U2> <int64> <int64> <int64>
19:18:59 Out [3]
<Frame>
<Index>                        a       b       <<U1>
<IndexHierarchy>
1                dd    0       1       1
<int64>          <<U2> <int64> <int64> <int64>

```
### New Behavior:
```python
<Series: (1, 'dd', 0)>
<Index>
a                      1
b                      1
<<U1>                  <int64>
<Frame>
<Index>                        a       b       <<U1>
<IndexHierarchy>
1                dd    0       1       1
<int64>          <<U2> <int64> <int64> <int64>
```
### Rationale:
The old implementation was inconsistent with other SF loc selections. The new impl follows the pattern of returning a series if given a single label, and a frame if given a list of labels (even with len==1)!
### Change 2
#### Setup
```python
frame1 = sf.Frame.from_dict_records(
        [dict(a=1,b=1), dict(a=2,b=3), dict(a=1,b=1), dict(a=2,b=3)],
        index=sf.IndexHierarchy.from_labels(
            [(1,'dd',0), (1,'bb',0), (2,'cc',0), (2,'ee',0)]
        )
)
frame2 = sf.Frame.from_dict_records(
        [dict(a=100,b=200), dict(a=20,b=30), dict(a=101,b=101), dict(a=201,b=301)],
        index=sf.IndexHierarchy.from_labels(
            [(1,'ddd',0), (1,'bbb',0), (2,'ccc',0), (2,'eee',0)]
        )
)
```
#### Old Behavior:
```python
with self.assertRaises(RuntimeError):
    sf.Frame.from_concat((frame1, frame2))
```
#### New Behavior:
```python
print(sf.Frame.from_concat((frame1, frame2)))
<Frame>
<Index>                        a       b       <<U1>
<IndexHierarchy>
1                dd    0       1       1
1                bb    0       2       3
2                cc    0       1       1
2                ee    0       2       3
1                ddd   0       100     200
1                bbb   0       20      30
2                ccc   0       101     101
2                eee   0       201     301
<int64>          <<U3> <int64> <int64> <int64>
```
#### Rationale
Not limited to indices that follow a strict tree hierarchy
### Change 3
#### Setup
```python
columns = sf.IndexHierarchy.from_product([10, 20], ['a', 'b'])
frame = sf.Frame(np.arange(8).reshape(2, 4), columns=columns)
frame = frame.drop[:]
print(frame.columns.dtypes)
```
#### Old Behavior:
```python
<Series>
<Index>
0        float64
1        float64
<int64>  <object>
```
#### New Behavior:
```python
<Series>
<Index>
0        int64
1        <U1
<int64>  <object>
```
#### Rationale:
The new implmentation will properly carry over the types from the original IndexHierarchy when drop[:] is called
# **sf.Series**
### Change 1
#### Setup
```python
s = sf.Series(range(5),
        index=sf.IndexHierarchy.from_labels(
            (('a', 'a'), ('a', 'b'), ('b', 'a'), ('b', 'b'), ('b', 'c'))
        )
)
```
#### Old Behavior:
```python
with self.assertRaises(RuntimeError):
    s.loc['a', :]
with self.assertRaises(RuntimeError):
    s.loc[['a', 'b'], 'b']
```
#### New Behavior:
```python
print(s.loc['a', :])
<Series>
<IndexHierarchy>
a                a     0
a                b     1
<<U1>            <<U1> <int64>
print(s.loc[['a', 'b'], 'b'])
<Series>
<IndexHierarchy>
a                b     1
b                b     3
<<U1>            <<U1> <int64>
```
#### Rationale
Old implementation forced all incoming selectors to either be a tuple or an HLoc. That restriction is removed, as comma seperated values will be parsed as a tuple
# **sf.IndexHierarchy**
### Change 1
#### No Setup
#### Old Behavior:
```python
with self.assertRaises(ErrorInitIndexLevel):
    ih = sf.IndexHierarchy.from_labels(tuple())
```
#### New Behavior:
```python
with self.assertRaises(ErrorInitIndex):
    ih = sf.IndexHierarchy.from_labels(tuple())
```
#### Rationale
Exception changes as there are no more `IndexLevel`s.
### Change 2
#### Setup
```python
ih = sf.IndexHierarchy.from_product(('I', 'II'), ('B', 'A'), (2, 1))

numbers = (n for n in range(1, 100000000))
def func(arg: tp.Any) -> int:
    return next(numbers)

print(ih.relabel_at_depth(func, depth_level=(l for l in (1, 0, 2))))
```
#### Old Behavior:
```python
<IndexHierarchyTree>
1                    3       5
1                    3       6
1                    4       7
1                    4       8
2                    9       11
2                    9       12
2                    10      13
2                    10      14
<int64>              <int64> <int64>
```
#### New Behavior:
```python
<IndexHierarchy>
1                3       5
1                3       6
1                4       5
1                4       6
2                3       5
2                3       6
2                4       5
2                4       6
<int64>          <int64> <int64>
```
#### Rationale
A side effect of the new approach is that relabel functions are only called for each unique label per depth, as opposed to on every match.
### Change 3
#### Setup
```python
labels = (
        ('I', 'A', 1, False),
        ('I', 'B', 2, True),
        ('II', 'C', 3, False),
        ('II', 'C', 4, True),
)
ih = sf.IndexHierarchy.from_labels(labels)
```
#### Old Behavior:
```python
print(ih.level_drop(-2))
<IndexHierarchy>
I                A
I                B
II               C
<<U2>            <<U1>
```
#### New Behavior:
```python
with self.assertRaises(ErrorInitIndexNonUnique):
    ih.level_drop(-2)
```
#### Rationale
The old implementation does not return an intuitive result. The expected result *should* be:
```python
<IndexHierarchy>
I                A
I                B
II               C
II               C
<<U2>            <<U1>
```
but that is not unique, so the new implementation fails.

flexatone:
Many thanks for this clear summary of these behavioral changes. By far most I see as improvements and/or corrections. 

One category, however, that should be considered further is the acceptance of partial, or incomplete labels as labels. To be consistent with normal `.loc` selection,  a label must be a complete tuple. Thus, I do not think that `s.loc['a', :]` should be equivalent `to s.loc[sf.HLoc['a', :]]`, nor should `s.loc[['a', 'b'], 'b']` be equivalent to `s.loc[sf.HLoc[['a', 'b'], 'b']]`: within a `loc`, a comma delimits axis, not depth, regardless if the container is a `sf.Series` or an `sf.Frame`; only within an `HLoc` does a comma delimit depth. 

Similarly `f1.loc[sf.HLoc[(1, 'dd')]]` is not equivalent to `f1.loc[(1, 'dd', 0)]`, as the former is really just a shorthand for `f1.loc[sf.HLoc[(1, 'dd', :)]]`, which is a "multiple" selection that, in the example above, should indeed return a `sf.Frame` and not an `sf.Series`. Only the equivalent expressions `f1.loc[sf.HLoc[(1, 'dd', 0)]]` or `f1.loc[(1, 'dd', 0)]` should return a `sf.Series` in this example.

The key conceptual issue here is that, outside of an `sf.HLoc`, a comma always delimits axis. Further, unspecified depths in an `sf.HLoc` are always a null slice, and a slice always returns a multiple selection.

The motivation for this strict distinction is so that the behavior of `loc` is consistent between `sf.Series` and `sf.Frame`, that there is no ambiguity in usage of `loc` on `sf.Frame` where the index might be made of tuples, and because Python cannot distinguish between `f[0, 1]` and `f[(0, 1)]`: both present to `__getitem__` a tuple of two elements. 

As one example of why this is important, consider this case:

```
>>> f = sf.Frame(np.arange(4).reshape(2, 2), index=((0, 1), (1, 2)))
>>> f
<Frame>
<Index>  0       1       <int64>
<Index>
(0, 1)   0       1
(1, 2)   2       3
<object> <int64> <int64>

>>> f.loc[0, 1]
KeyError: 0

>>> f.loc[(0, 1), :]
<Series: (0, 1)>
<Index>
0                0
1                1
<int64>          <int64>
```
If `f.loc[0, 1]` is, on a `sf.Series`, considered a single axis selection of `(0, 1)` (as suggested in your examples above), then the above usage of ` f.loc[0, 1]` should work. It correctly raises a `KeyError` because we must interpret the comma within a `loc` as delimiting axis. If we want to specify a tuple selection on an axis, we must provide an explicit tuple, and in this case, a second null-slice argument. 

I hope this example helps; I can provide more examples if useful.

flexatone:
@chaburkland : very promising results! 

Did you see my comments above regarding "partial" selection and the necesity of `tuple` / `HLoc` distinctions?

chaburkland:
## Another behavior change

`IndexHierarchy.from_labels`

This no longer raises a `RuntimeError` when given `reorder_for_hierarchy` and `continuation_token` in conjunction.
This is because the reordering process happens after the labels are parsed instead of before, which is why this limitation existed in the first place.

Example:
```python
labels = [(0, 0), (None, 1), (1, 0), (0, 2), (1, 1)] 

print(sf.IndexHierarchy.from_labels(labels, continuation_token=None))
<IndexHierarchy>
0                0
0                1
1                0
0                2
1                1
<int64>          <int64>

print(sf.IndexHierarchy.from_labels(labels, continuation_token=None, reorder_for_hierarchy=True))
<IndexHierarchy>
0                0
0                1
0                2
1                0
1                1
<int64>          <int64>
```

chaburkland:
@flexatone This is probably not ready to merge yet, but I would like to go through an initial round of reviews, as all tests are passing, performance is higher than before, I'm at 100% test coverage for what I added, and quality checks out.

There are parts that are a bit sloppy, but I would really like the extra pair of eyes to look through everything to give me direction on what is best to tackle.

static_frame/core/index_hierarchy.py
```diff
@@ -140,202 +277,252 @@ def from_product(cls: tp.Type[IH],
...
+                )
+
+    @classmethod
+    def _from_array(cls: tp.Type[IH],
+            array: np.ndarray,
+            *,
+            name: NameType = None,
+            depth_reference: tp.Optional[DepthLevelSpecifier] = None,
+            index_constructors: IndexConstructors = None,
+            ) -> IH:
```
flexatone, line 391: Usage of this method is problematic, as there are few cases (I expect) when each depth is the same dtype; and callers might be tempted to use `.values` to call this type, which might in inadvertently coerce types. Are you sure we need it?

static_frame/core/index_hierarchy.py
```diff
@@ -140,202 +277,252 @@ def from_product(cls: tp.Type[IH],
...
+                label_row = next(labels_iter)
+            except StopIteration:
+                break
+
+            if len(label_row) != depth:
+                raise ErrorInitIndex('All labels must have the same depth.')
+
+        # Convert to numpy arrays
+        for i in range(depth):
+            indexers[i] = np.array(indexers[i], dtype=int)
```
flexatone, line 501:
```suggestion
            indexers[i] = np.array(indexers[i], dtype=DTYPE_INT_DEFAULT)
```

static_frame/core/index_hierarchy.py
```diff
@@ -351,40 +538,69 @@ def from_index_items(cls: tp.Type[IH],
...
-                targets=targets,
-                own_index=True,
-                depth_reference=2, # depth always 2
+
+            if index_inner is None:
+                # This is the first index. Create an IndexGO
+                index_inner = tp.cast(IndexGO, mutable_immutable_index_filter(False, index))
+                new_indexer = PositionsAllocator.get(len(index_inner))
+            else:
+                new_labels = index.difference(index_inner) # Retains order!
```
flexatone, line 556: I am not sure this always retains order...

static_frame/core/index_hierarchy.py
```diff
@@ -351,40 +538,69 @@ def from_index_items(cls: tp.Type[IH],
...
+
+                if new_labels.size:
+                    index_inner.extend(new_labels.values)
+
+                new_indexer = index.iter_label().apply(index_inner._loc_to_iloc)
```
flexatone, line 561: If this is called frequently, we might use a lower-level interface.

static_frame/core/index_hierarchy.py
```diff
@@ -351,40 +538,69 @@ def from_index_items(cls: tp.Type[IH],
...
+
+            indexers_inner.append(new_indexer)
+            repeats.append(len(index))
+
+        if not labels:
+            assert index_inner is None # sanity check
+            return cls._from_empty(EMPTY_TUPLE, name=name, depth_reference=2)
+
+        index_inner = mutable_immutable_index_filter(cls.STATIC, index_inner) # type: ignore
+        assert isinstance(index_inner, Index) # mypy
```
flexatone, line 571: I do not like adding `isinstance` for mypy... if the alternative is a `#type: ignore`, that is a fine option.

static_frame/core/index_hierarchy.py
```diff
@@ -456,147 +674,186 @@ def _from_type_blocks(cls: tp.Type[IH],
...
-            elif d == depth_max:
-                # if there are redundancies here they will be caught in index creation
-                current.append(v)
-            else:
-                # cannot happen with TypeBlocks
-                raise ErrorInitIndex('label exceeded expected depth', v) #pragma: no cover
-            # shared implementation with from_labels ---------------------------
+        def gen_columns() -> tp.Iterator[np.ndarray]:
+            for block in blocks:
+                yield block.values.ravel()
```
flexatone, line 682: I am confused why we need `ravel` here...

static_frame/core/index_hierarchy.py
```diff
@@ -456,147 +674,186 @@ def _from_type_blocks(cls: tp.Type[IH],
...
+            ) -> None:
+        '''
+        Determines whether or not `indexers` are unique
+        '''
+        expected_len = len(values)
+
+        if expected_len <= 1:
+            return # Not possible to be duplicated!
+
+        unique_len = len(np.unique(np.array(indexers).T, axis=0))
```
flexatone, line 724: We have alternatives to `np.unique` in `util` now that are all more efficient.

static_frame/core/index_hierarchy.py
```diff
@@ -456,147 +674,186 @@ def _from_type_blocks(cls: tp.Type[IH],
...
+                for idx in indices._indices
+            ]
+            self._indexers = list(indices._indexers)
+            self._name = indices._name
+            self._blocks = indices._blocks
+            self._recache = False
+            self._index_types = None
+            return
+
+        if not all(isinstance(arr, np.ndarray) for arr in indexers):
```
flexatone, line 790:
```suggestion
        if not all(arr.__class__ is np.ndarray for arr in indexers):
```

static_frame/core/index_hierarchy.py
```diff
@@ -605,62 +862,69 @@ def copy(self: IH) -> IH:
...
         else:
-            if isinstance(depth_level, INT_TYPES):
-                yield from self._blocks._extract_array(column_key=depth_level)
-            else:
-                yield from array2d_to_tuples(
-                        self._blocks._extract_array(column_key=depth_level)
-                        )
+            yield from array2d_to_tuples(
+                    self._blocks._extract_array(column_key=depth_level)
+                    )
```
flexatone, line 916: Using `_extract_array` here may force undesirable type coercion; better to `zip` calls to `self.values_at_depth`, or find an alternative way to build these tuples.

static_frame/core/index_hierarchy.py
```diff
@@ -808,53 +1041,44 @@ def ndim(self) -> int:
...
+    def nbytes(self: IH) -> int:
         '''
         Return the total bytes of the underlying NumPy array.
 
         Returns:
             :obj:`int`
         '''
-        if self._recache:
-            self._update_array_cache()
         return self._blocks.nbytes
```
flexatone, line 1061: Should we include `nbytes` from other arrays managed in the new `IndexHierarchy`?

static_frame/core/index_hierarchy.py
```diff
@@ -1123,154 +1418,405 @@ def relabel_at_depth(self,
...
+        Indexers could be [2, 0, 0, 2, 1]
 
-    def _loc_to_iloc(self,
-            key: tp.Union[GetItemKeyType, HLoc]
-            ) -> GetItemKeyType:
+        This function return [C, A, B] -- shoutout to my initials
+        '''
+        # get the outer level, or just the unique frame labels needed
+        labels = self.values_at_depth(0)
+        label_indexes = sorted(ufunc_unique1d_positions(labels)[0])
```
flexatone, line 1511: Can we use a NumPy sort here? Does sort stability matter?

static_frame/core/index_hierarchy.py
```diff
@@ -1794,70 +2364,173 @@ class IndexHierarchyGO(IndexHierarchy):
...
+                label_index = self._indices[depth]._loc_to_iloc(label_at_depth)
+            else:
+                label_index = len(self._indices[depth])
+                self._indices[depth].append(label_at_depth)
 
-    def extend(self, other: IndexHierarchy) -> None:
+            label_indexers.append(label_index)
+
+        for i, label_index in enumerate(label_indexers):
+            self._indexers[i] = np.append(self._indexers[i], label_index)
```
flexatone, line 2405: Up until now, `np.append` has never been used in StaticFrame and is an anti-pattern in some regards. This is equivalent to an np.concatenate, and forces a fully copy of the underlying arrays for each addition. Please evaluate the performance of adding multiple single columns in a `FrameGO` with `IndexHierarchyGO` as columns with this implementation. I wonder if a cached approach, using a mutable list that defers creating an array as we do in `IndexGO`, might be better.

static_frame/core/index_hierarchy.py
```diff
@@ -1794,70 +2364,173 @@ class IndexHierarchyGO(IndexHierarchy):
...
+
+                    new_indexer = np.hstack(
+                        (self._indexers[depth], other._indexers[depth])
+                    )
+                    new_indexer.flags.writeable = False
+                    self._indexers[depth] = new_indexer
+                    continue
+
+                # Same labels, but different order. We have to remap the indexers.
+                indexer_remap = other_index.iter_label().apply(self_index._loc_to_iloc)
```
flexatone, line 2457: As in other contexts, there is likely a lower-level interface to doing the `iter_label().apply()` here.

static_frame/core/index_hierarchy.py
```diff
@@ -1794,70 +2364,173 @@ class IndexHierarchyGO(IndexHierarchy):
...
+            difference = ~other_index.isin(intersection)
+            del intersection
+
+            self_index.extend(other_index[difference])
+
+            # We use the mask_flags to signal what was in the intersection and what was not
+            # We can't use any existing masks, since the sizes will be different
+            other_index_remapped = other_index.iter_label().apply(
+                self_index._loc_to_iloc
+            ) * mask_flags[difference.astype(int)]
```
flexatone, line 2475:
```suggestion
            ) * mask_flags[difference.astype(DTYPE_INT_DEFAULT)]
```

flexatone (changes requested):
Thank you for this very significant reimplementation!

Lets call this a first-pass review! I was not able to review all of `index_hierarchy.py`, but wanted to get you some feedback now. I look forward to reviewing again after you have looked at my first round of comments.

static_frame/core/container_util.py
```diff
@@ -782,10 +782,7 @@ def key_to_ascending_key(key: GetItemKeyType, size: int) -> GetItemKeyType:
 def rehierarch_from_type_blocks(*,
         labels: 'TypeBlocks',
         depth_map: tp.Sequence[int],
-        index_cls: tp.Type['IndexHierarchy'],
-        index_constructors: tp.Optional[IndexConstructors] = None,
-        name: tp.Optional[tp.Hashable] = None,
-        ) -> tp.Tuple['IndexBase', np.ndarray]:
+        ) -> tp.Tuple["TypeBlocks", np.ndarray]:
```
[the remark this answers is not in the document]
chaburkland: Yeah - I decomposed this method to only return typeblocks, not an index. It's the callers job to create an index if that's what they need

static_frame/core/container_util.py
```diff
@@ -1106,7 +1098,7 @@ def _index_many_to_one(
     # if IndexHierarchy, collect index_types generators
     if index.ndim == 2:
         depth_first = index.depth
-        index_types_gen = [index._levels.index_types()] #type: ignore
+        index_types_gen = [index.index_types.values.tolist()]
```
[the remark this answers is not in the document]
chaburkland: Oh good point. I'll do this

static_frame/core/index.py
```diff
@@ -1226,17 +1226,19 @@ def level_add(self,
 
         if index_constructor is None:
             # cannot assume new depth is the same index subclass
-            index_constructors = (cls_depth, self.__class__)
-        else:
-            index_constructors = (index_constructor, self.__class__) #type: ignore
+            index_constructor = cls_depth
+
+        indices = [index_constructor((level,)), self.copy()]
```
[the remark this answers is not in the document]
chaburkland: Oh good point - this should use `immutable_index_filter(self)`! Since we are returning a new `IndexHierarchy` object, all it's internally managed indices must _not_ be `IndexGO` objects.

static_frame/core/index_base.py
```diff
@@ -120,20 +120,47 @@ def from_pandas(cls,
...
+                # a DatetimeIndex. Instead, it will be a regular Index with dtype=object.
+                # Only numpy datetime objects are put into a DatetimeIndex.
+                if isinstance(pd_idx, pandas.DatetimeIndex):
+                    constructor: tp.Type[Index] = IndexNanosecond
+                else:
+                    constructor = Index
+
+                if cls.STATIC:
+                    return constructor(pd_idx, name=pd_idx.name)
+                return tp.cast(Index, constructor._MUTABLE_CONSTRUCTOR(pd_idx, name=pd_idx.name))
```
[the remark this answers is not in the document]
chaburkland: Ah good point - I'll remove this.

static_frame/core/series.py
```diff
@@ -2802,7 +2803,7 @@ def to_pairs(self) -> tp.Iterable[tp.Tuple[tp.Hashable, tp.Any]]:
         Returns:
             tp.Iterable[tp.Tuple[tp.Hashable, tp.Any]]
         '''
-        if isinstance(self._index, IndexHierarchy):
+        if isinstance(self._index, (IndexHierarchy, IndexHierarchyTree)):
             index_values = array2d_to_tuples(self._index.values)
```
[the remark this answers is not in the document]
chaburkland: Sounds good :+1:

static_frame/core/index_hierarchy.py
chaburkland: This is a move I have to do alot. Let me explore a lower-level impl!

static_frame/core/index_hierarchy.py
chaburkland: Okay I will remove

static_frame/core/index_hierarchy.py
chaburkland: That seems reasonable. I'll add it!

static_frame/core/index_hierarchy.py
chaburkland: I was actually talking with Brandt this morning and learned for the first time what the purpose of `_recache`` and `_update_array_cache` was used for!

I will spend time thinking about how to implement this here :+1:

static_frame/core/index_hierarchy.py
chaburkland: I just implemented this in `Index._index_iloc_map`

static_frame/core/container_util.py
```diff
@@ -819,33 +816,28 @@ def rehierarch_from_type_blocks(*,
...
     labels_post = labels_post._extract(row_key=order_lex)
 
-    index = index_cls._from_type_blocks(
-            blocks=labels_post,
-            index_constructors=index_constructors,
-            name=name,
-            own_blocks=True,
-            )
-    return index, order_lex
+    return labels_post, order_lex
```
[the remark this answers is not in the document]
flexatone: Got it; I thought we were changing the type of the variable but I see we are not!

static_frame/core/loc_map.py
```diff
@@ -92,9 +92,11 @@ def map_slice_args(
                         pos += offset #type: ignore
                 else: # step
                     pos = attr # should be an integer
+                    if not isinstance(pos, int):
+                        raise TypeError(f'Step must be an integer, not {pos}')
```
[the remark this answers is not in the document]
flexatone: As long as we have a test that provides coverage I suppose it is fine... I just want to avoid `isinstance` wherever possible!

static_frame/core/type_blocks.py
```diff
@@ -2612,6 +2611,9 @@ def drop(self, key: GetItemKeyTypeCompound) -> 'TypeBlocks':
                 shape_reference=self._shape
                 )
 
+    def __iter__(self) -> tp.Iterator['TypeBlocks']:
+        for i in range(self.shape[1]):
+            yield self._extract(column_key=i)
```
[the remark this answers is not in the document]
flexatone: I see. I dislike added code just to make MyPy happy, and in this case, as I point out, there are two sensible options here as to whether an `np.ndarray` or a `TypeBlocks` instance should be extracted. But if Python simply does the same thing under `iter` without implementing this, I favor not implementing this function and silencing MyPy's complaint.

static_frame/core/index_hierarchy.py
```diff
@@ -72,61 +70,200 @@
...
+        array([0, 1, 2, 0, 1, 2, 0, 1, 2]),
+    ]
+
+    >>> lists = [[1, 2, 3], [4, 5, 6], [7, 8, 9]]
+    >>> build_indexers_from_product(tuple(map(len, lists)))
+    [
+        array([0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 2]),
+        array([0, 0, 0, 1, 1, 1, 0, 0, 0, 1, 1, 1, 0, 0, 0, 1, 1, 1]),
+        array([0, 1, 2, 0, 1, 2, 0, 1, 2, 0, 1, 2, 0, 1, 2, 0, 1, 2]),
+    ]
```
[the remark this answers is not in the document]
flexatone: Sure, that is reasonable.

static_frame/core/type_blocks.py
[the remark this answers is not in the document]
chaburkland: Isn't it better to be explicit though, instead of relying on default (and potentially confusing & opaque) behavior?

Or maybe to combine with your initial suggestion, should we make `__iter__`'s defintion raise `NotImplementedError`, and instead make an explicit method called `iter_array`?

static_frame/core/index_hierarchy.py
```diff
@@ -72,61 +70,200 @@
...
+        # Repeat each section `index_reps` times
+        indexer = np.tile(subsection, reps=group_reps)
+        indexer.flags.writeable = False
+        result.append(indexer)
+
+    return result
+
+
+def blocks_to_container(blocks: tp.Iterator[np.ndarray]) -> np.ndarray:
+    return TypeBlocks.from_blocks(blocks).values
```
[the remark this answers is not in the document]
flexatone: I see; this is used only in the `via_*` interfaces where we expect to get back blocks of uniform type anyways, and want to return an array. Best would be to re-implement implement this without using `TypeBlocks`; it would require realizing the iterator as a list so that two passe could be performed: the first to evaluate type and final size, the second to populate a new array. This is out-of-scope perhaps of this PR, so maybe just create an issue for future enhancement.

static_frame/core/type_blocks.py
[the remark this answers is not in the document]
flexatone: Raising `NotImplemented()` is attractive. I do not think we need an `iter_array`, so perhaps nothing more to do at this point.

static_frame/core/index_hierarchy.py
flexatone: haha, lets hope I did not!

static_frame/core/index_hierarchy.py
flexatone: I see. Is `column_1d_filter` better... or if you know it is `(n, 1)` you can just directly reshape. `ravel` is one of those "handle all cases" NumPy functions I try to avoid...

static_frame/core/index_hierarchy.py
flexatone: You mean in a sorted order, not the order they appear? As `ufunc_unique1d_positions` hands back arrays, however, I still do not see why we need to create a Python list and sort it, rather than just sorting the NumPy array.

chaburkland: Oh sorry - I misread your original comment. Yes I will use `np.sort`!

static_frame/core/type_blocks.py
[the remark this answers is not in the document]
chaburkland: What do you mean we mean we don't need `iter_array`? There are multiple places throughout the code base that `for block in blocks` is called.

If we don't want to define `__iter__` for the ambiguity you've mentioned, I don't know how not defining it, but relying on the default implementation addresses the issue of "What does `for block in blocks` even mean?"

My though is to set `__iter__` to raise `NotImplementedError` (not `NotImplemented`), and instead replace all instances throughout the code base with `for block in blocks.iter_array()`.

static_frame/core/index_hierarchy.py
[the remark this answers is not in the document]
chaburkland: https://github.com/InvestmentSystems/static-frame/issues/443

static_frame/core/type_blocks.py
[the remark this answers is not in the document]
flexatone: Are you sure that, in the places we are calling `for block in blocks`, `blocks` is a `TypeBlocks`? What I see as far more common is that `blocks`, in those scenarios, is a `tp.Iterable[np.ndarray]` (which we often refer to as "blocks"), not `TypeBlocks`.

I am not opposed to implementing `iter_array`, I just suspect we do not need it. If you have a case in the `IndexHierarchy` implementation, then that might be good enough reason.

And yes, `NotImplementedError` not `NotImplemented`!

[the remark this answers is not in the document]
chaburkland: Oooooo I see now. Okay yeah no one is actually iterating over a `TypeBlocks` instance. I will raise an error, and just implement something explicit in `IndexHierarchy`

chaburkland:
@flexatone 

Either you need to re-review this, or someone needs to sedate me. I am stuck in an endless loop of continually cleaning up and tightening the logic and I fear I cannot stop until this is merged.

flexatone:
Thank you, @chaburkland ! I will review this one more time by Friday, and we can merge it in early next week; I might do one more release this week before we start on the 0.9 branch.

static_frame/core/index_hierarchy.py
```diff
@@ -456,211 +719,273 @@ def _from_type_blocks(cls: tp.Type[IH],
...
+            if indices._recache:
+                indices._update_array_cache() # pragma: no cover
+
+            self._indices = [
+                mutable_immutable_index_filter(self.STATIC, index)
+                for index in indices._indices
+            ]
+            self._indexers = indices._indexers
+            self._name = name if name is not NAME_DEFAULT else indices._name
+            self._blocks = indices._blocks
```
flexatone, line 814: Are we sure that we can take a reference here and not a shallow copy? I.e., is `self._blocks` never mutated in place?

static_frame/core/index_hierarchy.py
```diff
@@ -456,211 +719,273 @@ def _from_type_blocks(cls: tp.Type[IH],
...
-            raise NotImplementedError(f'no handling for creation from {levels}')
+            self._blocks = self._create_blocks_from_self()
 
-        if self.STATIC and index_level.STATIC:
-            self._levels = index_level
-        else: # must deepcopy IndexLevels if not IndexHierarchy not static
-            self._levels = index_level.to_index_level(
-                    cls=self._LEVEL_CONSTRUCTOR
-                    )
+        self._values = self._blocks.values
```
flexatone, line 843: This will force type consolidation: is that acceptable?

static_frame/core/index_hierarchy.py
```diff
@@ -456,211 +719,273 @@ def _from_type_blocks(cls: tp.Type[IH],
...
 
-    #---------------------------------------------------------------------------
-    def __deepcopy__(self: IH, memo: tp.Dict[int, tp.Any]) -> IH:
+        current_size = len(self._blocks)
+
+        for depth, indexer in enumerate(self._indexers):
+            new_indexers[depth][:current_size] = indexer
+
+        # TODO-Ariza: Should we clear self._indexers?
+        # TODO-Ariza: What if an error is raised anywhere onward in this method?
```
flexatone, line 858: We want to avoid malformed instances being created / retained, but not at any material memory cost.

static_frame/core/index_hierarchy.py
```diff
@@ -456,211 +719,273 @@ def _from_type_blocks(cls: tp.Type[IH],
...
+        new_indexers = [np.empty(self.__len__(), DTYPE_INT_DEFAULT) for _ in range(self.depth)]
 
-    #---------------------------------------------------------------------------
-    def __deepcopy__(self: IH, memo: tp.Dict[int, tp.Any]) -> IH:
+        current_size = len(self._blocks)
+
+        for depth, indexer in enumerate(self._indexers):
+            new_indexers[depth][:current_size] = indexer
+
+        # TODO-Ariza: Should we clear self._indexers?
```
flexatone, line 857: If that means eagerly clearing memory in this method, might as well.

static_frame/core/index_hierarchy.py
```diff
@@ -456,211 +719,273 @@ def _from_type_blocks(cls: tp.Type[IH],
...
+        # TODO-Ariza: What if an error is raised anywhere onward in this method?
+
+        offset = current_size
+
+        # For all these extensions, we have already update self._indices - we now need to map indexers
+        for pending in self._pending_extensions: # pylint: disable = E1133
+            if isinstance(pending, PendingRow):
```
flexatone, line 864: No chance of subclass, right?
```suggestion
            if pending.__class__ is PendingRow:
```

static_frame/core/index_hierarchy.py
```diff
@@ -456,211 +719,273 @@ def _from_type_blocks(cls: tp.Type[IH],
...
-                    cls=self._LEVEL_CONSTRUCTOR
-                    )
+        self._values = self._blocks.values
+        self._map = HierarchicalLocMap(indices=self._indices, indexers=self._indexers)
 
-        if self._levels.depth <= 1:
-            raise ErrorInitIndex(f'invalid depth ({self._levels.depth}) for IndexLevels composed in IndexHierarchy')
+    def _update_array_cache(self: IH) -> None:
+        # This MUST be set before entering this context
+        assert self._pending_extensions is not None
```
flexatone, line 848: Is this still needed?

static_frame/core/index_hierarchy.py
```diff
@@ -456,211 +719,273 @@ def _from_type_blocks(cls: tp.Type[IH],
...
+
+    # --------------------------------------------------------------------------
+
+    def __setstate__(self, state: tp.Tuple[None, tp.Dict[str, tp.Any]]) -> None:
+        '''
+        Ensure that reanimated NP arrays are set not writeable.
+        '''
+        for key, value in state[1].items():
+            setattr(self, key, value)
+        self._values.flags.writeable = False
```
flexatone, line 901: Does `writeable` need to be set for `self._indexers`?

static_frame/core/index_hierarchy.py
```diff
@@ -829,32 +1139,40 @@ def nbytes(self) -> int:
         '''
         if self._recache:
             self._update_array_cache()
+
+        total = sum(map(_NBYTES_GETTER, self._indices))
+        total += sum(map(_NBYTES_GETTER, self._indexers))
+        total += self._blocks.nbytes
+        total += self._map.nbytes
```
flexatone, line 1146: Do you need to return `total`?

static_frame/core/index_hierarchy.py
```diff
@@ -949,59 +1275,77 @@ def _ufunc_set(self,
...
+
+    def _drop_iloc(self: IH,
+            key: GetItemKeyType,
+            ) -> IH:
+        '''
+        Create a new index after removing the values specified by the iloc key.
         '''
         if self._recache:
-            self._update_array_cache()
+            self._update_array_cache() # pragma: no cover
```
flexatone, line 1301: Why not cover this scenario with a test?

static_frame/core/index_hierarchy.py
```diff
@@ -1018,259 +1362,580 @@ def values_at_depth(self,
...
+            self._update_array_cache()
+
+        # Could this be memoized?
+        def _extract_counts(
+                arr: np.ndarray,
+                pos: int,
+                ) -> tp.Iterator[tp.Tuple[tp.Hashable, int]]:
+            unique, widths = ufunc_unique1d_counts(arr)
+            labels = self._indices[pos].values[unique]
+            yield from zip(labels, widths)
```
flexatone, line 1403: Should this just be a class or module-level function?

static_frame/core/index_hierarchy.py
```diff
@@ -1018,259 +1362,580 @@ def values_at_depth(self,
...
+            return
+
+        def gen()-> tp.Iterator[tp.Tuple[tp.Hashable, int]]:
+            # We need to build up masks for each depth level. This requires the combination of
+            # all depths above it. We do not care about order since we are only determining
+            # counts. As such, we simply walk from 1 -> len(level) for each outer level
+            ranges = tuple(map(range, map(len, self._indices[:pos])))
+
+            for outer_level_idxs in itertools.product(*ranges):
+                screen = np.full(self.__len__(), True, dtype=bool)
```
flexatone, line 1418:
```suggestion
                screen = np.full(self.__len__(), True, dtype=DTYPE_BOOL)
```

static_frame/core/index_hierarchy.py
```diff
@@ -1018,259 +1362,580 @@ def values_at_depth(self,
...
+                new_label = get_new_label(label)
+
+                if new_label not in new_index:
+                    new_index[new_label] = len(new_index)
+                else:
+                    index_remap[label_idx] = new_index[new_label]
+
+            new_indices[level] = index.__class__(new_index)
+
+            indexer = np.array([index_remap.get(i, i) for i in self._indexers[level]])
```
flexatone, line 1586: Can we provide a dtype arg here?

static_frame/core/index_hierarchy.py
```diff
@@ -1018,259 +1362,580 @@ def values_at_depth(self,
...
+        # Key is already a mask!
+        if key_at_depth.__class__ is np.ndarray and key_at_depth.dtype == DTYPE_BOOL: # type: ignore
+            return key_at_depth
+
+        index_at_depth = self._indices[depth]
+        indexer_at_depth = self._indexers[depth]
+
+        if isinstance(key_at_depth, slice):
+            if key_at_depth.start is not None:
+                start: int = tp.cast(int, index_at_depth.loc_to_iloc(key_at_depth.start))
```
flexatone, line 1653: Again, I am cautious in using `tp.cast` in anything that might be called very frequently.

static_frame/core/index_hierarchy.py
```diff
@@ -1018,259 +1362,580 @@ def values_at_depth(self,
...
+        for key_index, self_index, key_indexer in zip(
+                key._indices,
+                self._indices,
+                key._indexers,
+                ):
+            indexer_remap = key_index._index_iloc_map(self_index)
+            key_indexers.append(indexer_remap[key_indexer])
 
-        >>> index.relabel_at_depth(mapper, depth_level=[0, 1, 2])
+        key_indexers = np.array(key_indexers).T
```
flexatone, line 1715: Can we specify a `dtype` in this `np.array` call?

static_frame/core/index_hierarchy.py
```diff
@@ -1282,92 +1947,100 @@ def _ufunc_binary_operator(self, *,
...
+
+        if self._recache:
+            self._update_array_cache() # pragma: no cover
 
         if isinstance(other, Index):
             other = other.values
         elif isinstance(other, IndexHierarchy):
             if other._recache:
-                other._update_array_cache()
+                other._update_array_cache() # pragma: no cover
```
flexatone, line 1965: Why skip coverage of this scenario?

static_frame/core/index_hierarchy.py
```diff
@@ -1282,92 +1947,100 @@ def _ufunc_binary_operator(self, *,
...
-        return self._levels.__contains__(value)
-    #---------------------------------------------------------------------------
+        Determine if a label `value` is contained in this Index.
+        '''
+        try:
+            result = self._loc_to_iloc(value)
+        except KeyError:
+            return False
+
+        if isinstance(result, (np.ndarray, list)):
```
flexatone, line 2034: Can this be:
```suggestion
        if hasattr(result, '__len__'):
```

static_frame/core/index_hierarchy.py
```diff
@@ -1413,67 +2082,83 @@ def equals(self,
...
+        Return a new Index with the labels sorted.
 
         Args:
             {ascendings}
             {kind}
             {key}
         '''
         if self._recache:
-            self._update_array_cache()
+            self._update_array_cache() # pragma: no cover
```
flexatone, line 2137: Why skip coverage of this scenario?

static_frame/core/index_hierarchy.py
```diff
@@ -1485,26 +2170,32 @@ def isin(self, other: tp.Iterable[tp.Iterable[tp.Hashable]]) -> np.ndarray:
...
-        '''Return an :obj:`IndexHierarchy` with values rotated forward and wrapped around (with a positive shift) or backward and wrapped around (with a negative shift).
+    def roll(self: IH,
+            shift: int,
+            ) -> IH:
+        '''
+        Return an :obj:`IndexHierarchy` with values rotated forward and wrapped around (with a positive shift) or backward and wrapped around (with a negative shift).
         '''
         if self._recache:
-            self._update_array_cache()
+            self._update_array_cache() # pragma: no cover
```
flexatone, line 2180: Why skip coverage of this scenario?

static_frame/core/index_hierarchy.py
```diff
@@ -1513,42 +2204,46 @@ def fillna(self, value: tp.Any) -> 'IndexHierarchy':
...
+        '''
+        Selects a deterministically random sample from this IndexHierarchy.
+
+        Returns:
+            The sampled IndexHierarchy
+            An integer array of sampled iloc values
+        '''
         if self._recache:
-            self._update_array_cache()
+            self._update_array_cache() # pragma: no cover
```
flexatone, line 2228: Same here...

static_frame/core/index_hierarchy.py
```diff
@@ -1563,33 +2258,40 @@ def iloc_searchsorted(self,
...
+            )
 
-        dt_pos = np.array([issubclass(idx_type, IndexDatetime)
-                for idx_type in self._levels.index_types()])
+        dt_pos = np.array(
+            [
+                issubclass(idx_type, IndexDatetime)
+                for idx_type in self._index_constructors
+            ]
+        )
```
flexatone, line 2270:
```suggestion
        dt_pos = np.array(
            [
                issubclass(idx_type, IndexDatetime)
                for idx_type in self._index_constructors
            ],
            dtype=DTYPE_BOOL,
        )
```

static_frame/core/index_hierarchy.py
```diff
@@ -1603,184 +2305,209 @@ def loc_searchsorted(self,
...
 
-    def _to_frame(self,
-            constructor: tp.Type['Frame']
+    def _to_frame(self: IH,
+            constructor: tp.Type['Frame'],
             ) -> 'Frame':
-
         if self._recache:
-            self._update_array_cache()
+            self._update_array_cache() # pragma: no cover
```
flexatone, line 2329: Same here...

static_frame/core/index_hierarchy.py
```diff
@@ -1603,184 +2305,209 @@ def loc_searchsorted(self,
...
+
+        return tree
+
+    def to_tree(self: IH) -> TreeNodeT:
         '''
-        return self._levels.traverse() # type: ignore
+        Returns the tree representation of an IndexHierarchy
+        '''
+        if self._recache:
+            self._update_array_cache() # pragma: no cover
```
flexatone, line 2401: Why no coverage?

static_frame/core/index_hierarchy.py
```diff
@@ -1603,184 +2305,209 @@ def loc_searchsorted(self,
...
-            levels_src = self._levels.to_index_level()
+            indices = [index_cls((level,)), *(idx.copy() for idx in self._indices)]
 
-        index_cls = self._INDEX_CONSTRUCTOR if index_constructor is None else index_constructor
+        indexers = np.array(
+                [
+                    np.zeros(self.__len__(), dtype=DTYPE_INT_DEFAULT),
+                    *self._indexers
+                ]
+        )
```
flexatone, line 2438:
```suggestion
        indexers = np.array(
                [
                    np.zeros(self.__len__(), dtype=DTYPE_INT_DEFAULT),
                    *self._indexers
                ],
                dtype=DTYPE_INT_DEFAULT,
        )
```

static_frame/core/index_hierarchy.py
```diff
@@ -1603,184 +2305,209 @@ def loc_searchsorted(self,
...
+            ) -> tp.Union[Index, IH]:
+        '''
+        Return an IndexHierarchy with one or more leaf levels removed.
 
         Args:
             count: A positive value is the number of depths to remove from the root (outer) side of the hierarchy; a negative value is the number of depths to remove from the leaf (inner) side of the hierarchy.
         '''
         if self._recache:
-            self._update_array_cache()
+            self._update_array_cache() # pragma: no cover
```
flexatone, line 2464: Why no coverage?

static_frame/core/index_hierarchy.py
```diff
@@ -1794,69 +2521,89 @@ class IndexHierarchyGO(IndexHierarchy):
...
+            ) -> IndexHierarchy:
+        '''
+        Entrypoint to `astype` the container
+        '''
         from static_frame.core.index_datetime import dtype_to_index_cls
         container = self.container
 
         if container._recache:
-            container._update_array_cache()
+            container._update_array_cache() # pragma: no cover
```
flexatone, line 2606: Why no coverage?

flexatone (changes requested):
Excellent work with this! A few small optimizations as well as some additional coverage, and this will be ready for 0.9!

static_frame/core/index_hierarchy.py
chaburkland: Whoops!

static_frame/core/index_hierarchy.py
chaburkland: Okay I'll `# type: ignore`

static_frame/core/index_hierarchy.py
chaburkland: I've made `_indexers` non-writeable everywhere. I'm not necessarily concerned about someone doing it, but I feel like it's better to ensure state doesn't change wherever possible

static_frame/core/index_hierarchy.py
chaburkland: I don't believe it's possible without material memory overhead, so I will resolve this

static_frame/core/index.py
```diff
@@ -726,6 +727,37 @@ def positions(self) -> np.ndarray:
...
+        '''
+        if self.__len__() == 0:
+            return EMPTY_ARRAY
+
+        ar1 = self.values
+        ar2 = other.values
+
+        ar1, ar1_indexer = ufunc_unique1d_indexer(ar1)
+
+        aux = np.concatenate((ar1, ar2))
```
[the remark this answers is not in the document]
chaburkland: Good catch! This does not handle types. Let me use `concat_resolved`

static_frame/core/loc_map.py
```diff
@@ -167,23 +171,214 @@ def loc_to_iloc(cls, *,
...
+
+        is_single_key = True
+
+        # 1. Perform label resolution
+        for key_at_depth, index_at_depth in zip(key, indices):
+            if self.is_single_element(key_at_depth):
+                key_indexers.append((index_at_depth._loc_to_iloc(key_at_depth),))
+            else:
+                is_single_key = False
+                subkey_indexers: tp.List[int] = []
```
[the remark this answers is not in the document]
chaburkland: Why? This variable only exists in side the `else` clause inside the `for-loop`. And typehints are ignored at runtime. I think putting it anywhere else would harm code readability

static_frame/core/index_hierarchy.py
```diff
@@ -220,122 +467,108 @@ def from_labels(cls: tp.Type[IH],
...
+        depth = len(label_row)
+
+        if depth == 1:
+            raise ErrorInitIndex('Cannot create IndexHierarchy from only one level.')
+
+        # A mapping for each depth level, of label to index
+        hash_maps: tp.List[tp.Dict[tp.Hashable, int]] = [{} for _ in range(depth)]
+        indexers: tp.List[tp.List[int]] = [[] for _ in range(depth)]
+
+        prev_row: tp.Sequence[tp.Hashable] = EMPTY_TUPLE
```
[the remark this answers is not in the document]
chaburkland: I lean towards no, since these types are only applicable to this specific context. I think it would harm readability to put these anywhere else.

static_frame/core/index_hierarchy.py
chaburkland: Yeah - this is only used by the `values` property.

static_frame/core/index_hierarchy.py
chaburkland: I got tired of adding entire tests just to add coverage to these.

If I wasn't so concerned about performance, I would have designed it like this:

```python
def _update_array_cache(self):
    if not self._recache:
        return
    ...

def meth1(self):
    self._update_array_cache()
    ...

def meth2(self):
    self._update_array_cache()
    ...
```
However, since this is _slightly_ slower than the other approach (perform the boolean check _before_ the function call), it was not implemented this way.

An annoying side effect is that now it requires a how lot of redundant tests just to capture the branch, which after doing for a while, I got tired of and so I just went with `pragma: no-cover`

static_frame/core/index_hierarchy.py
chaburkland: It needs `self._indices`!

static_frame/core/index_hierarchy.py
chaburkland: See this response: https://github.com/InvestmentSystems/static-frame/pull/437#discussion_r831501537

static_frame/core/index_hierarchy.py
chaburkland: See this response: https://github.com/InvestmentSystems/static-frame/pull/437#discussion_r831501537

static_frame/core/index_hierarchy.py
chaburkland: See this response: https://github.com/InvestmentSystems/static-frame/pull/437#discussion_r831501537

static_frame/core/index_hierarchy.py
chaburkland: See this response: https://github.com/InvestmentSystems/static-frame/pull/437#discussion_r831501537

static_frame/core/index_hierarchy.py
chaburkland: See this response: https://github.com/InvestmentSystems/static-frame/pull/437#discussion_r831501537

static_frame/core/index_hierarchy.py
chaburkland: See this response: https://github.com/InvestmentSystems/static-frame/pull/437#discussion_r831501537

static_frame/core/index_hierarchy.py
chaburkland: See this response: https://github.com/InvestmentSystems/static-frame/pull/437#discussion_r831501537

static_frame/core/index_hierarchy.py
chaburkland: See this response: https://github.com/InvestmentSystems/static-frame/pull/437#discussion_r831501537

static_frame/test/unit/test_index.py
```diff
@@ -591,8 +591,8 @@ def test_index_go_f(self) -> None:
         idx1 = IndexAutoFactory.from_optional_constructor(3,
                 default_constructor=IndexGO)
         idx1.append(3) # type: ignore
-        post = idx1._loc_to_iloc(np.array([True, False, True, False]), 2) #type: ignore
-        self.assertEqual(post.tolist(), [2, 4]) #type: ignore
+        post = idx1._loc_to_iloc(np.array([True, False, True, False]))
+        self.assertEqual(post.tolist(), [True, False, True, False]) #type: ignore
```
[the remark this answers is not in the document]
chaburkland: Yeah `_loc_to_iloc` can return boolean arrays. It's actually more efficient too since it just passes through. Honestly I'm not sure what this old behavior was doing, as it relates to `IndexLevel._loc_to_iloc` and `offset`s

static_frame/core/index_hierarchy.py
chaburkland: This does apply here! Let met add this.

static_frame/core/loc_map.py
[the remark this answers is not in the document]
flexatone: While the application of type hints is ignored at runtime, it is my understanding that the objects are still evaluated and constructed, and creating those instances takes time. I would be happy to be wrong about this. Perhaps if the usage is between `:` and `=` is ignored? But in an interactive session we get syntax errors: 

```
>>> foo: tp.Blah = 30                                                                                                                                                                                                
AttributeError: module 'typing' has no attribute 'Blah'
```
And here we can see that complicated type hints incur greater creation cost:
```
>>> %timeit tp.List[int]                                                                                                                                                                                             
434 ns ± 34.6 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
>>> %timeit tp.List[tp.Dict[tp.Tuple[str, int], tp.Tuple[str, int]]]                                                                                                                                                 
3.27 µs ± 232 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

static_frame/core/index_hierarchy.py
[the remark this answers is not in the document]
flexatone: My suggestion here refers to my assumption above in regard to `tp` instance creation.

static_frame/core/index_hierarchy.py
flexatone: `TypeBlocks` is not immutable; only its underlying arrays are. What about in the case of an `IndexHierarchyGO`?

static_frame/core/index_hierarchy.py
flexatone: So we are eagerly-caching `values`; but is that necessary? Is it better to be lazy and/or not hold on to `values` and simply create `values` on the fly when requested (exactly what `TypeBlocks` does).

static_frame/core/index_hierarchy.py
flexatone: Sure, `assert` is fine for now.

static_frame/core/index_hierarchy.py
flexatone: Makes sense.

static_frame/core/loc_map.py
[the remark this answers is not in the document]
chaburkland: This is shocking information to me. If this is the case, the it almost seems like no code being heavily repeated should ever be typehinted!

static_frame/core/index_hierarchy.py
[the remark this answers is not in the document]
chaburkland: I see. Do you think the loss of readability is worth it, given that this is a nanosecond cost happening in an otherwise expensive constructor? The construction of typehints is effectively 0 when compared to the cost of `from_labels`.

static_frame/core/index_hierarchy.py
chaburkland: If the array is `IndexHierarchyGO`, then `_blocks` will get updated in `_update_array_cache`, which gets called right before this move, so we can safely create a reference to `indices._blocks`

static_frame/core/index_hierarchy.py
flexatone: I agree that using the in-function branch is best. But I do not think that these cases justify `pragma: no-cover`; the cases we use `pragma: no cover` are very few. You might find that you have put this line in functions that do not need it, or other situations that arise when you have deferred operations in this particular context.

static_frame/core/index_hierarchy.py
chaburkland: Initially I saw it as an opporunity for improvement, but I think that was before I fully realized that `values` isn't really a core feature of the class. I'll revert it back to be created when requested.

static_frame/core/loc_map.py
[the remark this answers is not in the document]
flexatone: Yes... or the typehints need to be defined outside the function and used as a singleton!

static_frame/core/index_hierarchy.py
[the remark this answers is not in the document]
flexatone: I think the readability is improved!
```python
HasableToIntMaps = tp.List[tp.Dict[tp.Hashable, int]]
```

static_frame/core/loc_map.py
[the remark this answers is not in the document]
chaburkland: But even the singleton requires a `globals()` lookup which is still overhead.

static_frame/core/index_hierarchy.py
flexatone: Sorry, even better:
```python
        dt_pos = np.fromiter(
            (
                issubclass(idx_type, IndexDatetime)
                for idx_type in self._index_constructors
            ),
            count=len(self._index_constructors),
            dtype=DTYPE_BOOL,
        )
```

static_frame/core/index_hierarchy.py
chaburkland: :cry:

static_frame/core/loc_map.py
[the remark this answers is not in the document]
flexatone: Right, but that is far better than creating four or five object instances!

chaburkland:
@flexatone I believe I have addressed everything. If you are good, I think it's probably time to merge

flexatone (approved):
Thanks again for this excellent contribution.

Issue #455, opened 2022-04-08
IndexHierarchy selection from IndexHierarchy does not account for ordering


chaburkland:
**Description**

When selecting an `IndexHierarchy` selection from another `IndexHierarchy`, the order of the selection is ignored.

**Example**

```python
>>> a = sf.IndexHierarchy.from_labels([(0, 1), (0, 2), (1, 0), (1, 1)])
>>> b = sf.IndexHierarchy.from_labels([(1, 1), (0, 2), (0, 1)]) 

>>> tuple(b)
((1, 1), (0, 2), (0, 1))
>>> tuple(a.loc[b])
((0, 1), (0, 2), (1, 1))
```

flexatone:
This issue was corrected with #456 and will be part of the 0.9.2 release.

PR #456, opened 2022-04-08
Fixes bug (plus a bonus one!) and adds tests.

chaburkland:
Refs: #455


flexatone (approved):
Nice work! Please update new.rst in regard to this fix and I will get a new release out shortly.

PR #458, opened 2022-04-10
Actually fixes the bug I thought I fixed before.

chaburkland:


static_frame/core/index_hierarchy.py
```diff
@@ -1711,29 +1712,22 @@ def _loc_to_iloc_index_hierarchy(self: IH,
...
-                )
+        remapped_indexers = np.array(remapped_indexers, dtype=DTYPE_UINT_DEFAULT).T
 
-        if len(ilocs) != len(key):
-            raise KeyError(
-                f"Key contains labels that are not in this IndexHierarchy, including: {key._drop_iloc(ilocs)[0]}"
-            )
-
-        return ilocs[ilocs_order]
+        try:
```
chaburkland, line 1726: The tricky part of this function, is that it's not enough to remap.

For example, lets say `A` is:
```
0, 0
1, 1
```
and `B` is
```
1, 0
0, 1
```
The remapping would work, since both indexers share the same labels at each depth.

The problem is, that `A` doesn't contain the label `(1, 0)`. I originally addressed this by using `intersect1d`, but that led to the issue where order wasn't being retained. In the last fix attempt, I incorrectly assumed from one example that doing `ilocs[ilocs_order]` was enough to ensure the order was retained, but that was not correct.

I realized that the better approach would be to take advantage of `self._map`, which already has optimized infrastructure to map indexers to unique integer locations. So I could simply take the `remapped_indexers`, and then map them to iloc locations. If any of them are invalid, then the mapping will raise a `KeyError`.

static_frame/core/loc_map.py
```diff
@@ -394,3 +394,17 @@ def loc_to_iloc(self: _HLMap,
...
+        '''Modifies indexers in-place'''
+        # Validate input requirements
+        assert indexers.ndim == 2
+        assert indexers.shape[1] == len(self.bit_offset_encoders)
+        assert indexers.dtype == DTYPE_UINT_DEFAULT
+
+        indexers <<= self.bit_offset_encoders
+
+        indexers = np.bitwise_or.reduce(indexers, axis=1)
+        return list(map(self.encoded_indexer_map.__getitem__, indexers))
```
chaburkland, line 410: This is the same logic as `HierarchyLocMap.loc_to_iloc`, except that's operating on a 2D indexer array, instead of a single key that hasn't had label resolving applied yet, so the logic couldn't be cleanly abstracted into a single base function (as far as I could tell).

flexatone:
Thanks for this improvement! I see Codecov is saying one line is missing coverage: can you check?

PR #467, opened 2022-04-30
Improved name extraction in extracing `Series` from `Frame`

flexatone:


static_frame/core/frame.py
```diff
@@ -4313,10 +4313,8 @@ def _extract(self,
             index = self._index
         else:
             index = self._index._extract_iloc(row_key)
-            if not row_key_is_slice:
-                name_row = self._index.values[row_key]
-                if self._index.depth > 1:
-                    name_row = tuple(name_row)
+            if not row_key_is_slice and isinstance(row_key, INT_TYPES):
+                name_row = self._index._extract_iloc_by_int(row_key)
```
chaburkland, line 4317: Why not just:

`name_row = self.index._extract_iloc(row_key)`

`_extract_iloc_by_int` seems redundant, as `_extract_iloc` already supports integer keys.

chaburkland (commented):
The `index.values` call hurts for `IndexHierarchies`! :joy: I'm glad you found that!

static_frame/core/frame.py
flexatone: While `IndexHierarchy._extract_iloc()` is relatively simple for an `int`-based selection, `Index._extract_iloc()` does numerous checks before using the selector as an `int`. In both cases we do an `isinstance()` check. In the context the function here, we already have to do the `isinstance()` check before calling this function on the `IndexBase`, so doing an `isinstance` check immediately again seems unnecessary and wasteful. Thus, a specialized selector provides an optimal path. This is in-line with some new interfaces in `TypeBlocks` where we are implementing more narrow selectors for callers that know what type they have or exactly what they want, rather than always using the general-purpose, all-in-one interfaces we need for the public entry points.

PR #482, opened 2022-05-20
Improved tuple row iteration

flexatone:


static_frame/core/type_blocks.py
```diff
@@ -564,6 +564,20 @@ def mloc(self) -> np.ndarray:
     def unified(self) -> bool:
         return len(self._blocks) <= 1
 
+    @property
+    def unified_dtypes(self) -> bool:
```
chaburkland, line 568: Wouldn't this be a good candidate for caching?

flexatone: Well, `TypeBlocks` are mutable, so caching would not be trivial. If we had a `TypeBlocks` and a `TypeBlocksGO`, that might make it easier.

PR #483, opened 2022-05-20
`fill_value` to support mappings, `FillValueAuto` implementation

flexatone:


static_frame/core/fill_value_auto.py
```diff
@@ -6,34 +6,83 @@
 from static_frame.core.util import NAT
 from static_frame.core.util import NAT_TD64
 
+FILL_VALUE_AUTO_DEFAULT = object()
```
ForeverWintr, line 9: I notice this name is the same as `container_util.FILL_VALUE_AUTO_DEFAULT`, which could be confusing. Maybe call this `FILL_VALUE_UNSET` or something?

nekobon: I like `FILL_VALUE_UNSET`! I read the code before reading this comment and got confused.

flexatone: Great point; thanks!

PR #516, opened 2022-07-14
Optimizes how rehierarch_from_type_blocks is implemented. Fixes tests.

chaburkland:


flexatone (approved):
Looks good; just one suggestion.

flexatone:
I now recall that the old behavior, keeping labels in their first-observed order but making them hierarchable (in the old sense) was a feature. I am not sure that global sorting is what we always want to do. Lets consider that before merging this change.

chaburkland:
@flexatone I can't remember what we were doing with this PR. Did we come to any conclusions, namely, either merge, merge with changes, or abandon entirely?

chaburkland:
@flexatone I can't remember what we were doing with this PR. Did we come to any conclusions, namely, either merge, merge with changes, or abandon entirely?

flexatone:
@chaburkland : As discussed above, I think retaining ordering in rehierarching is important. Does this retain order, or do a full lex sort on all depths?

PR #523, opened 2022-07-19
Prevent unhandled exception in ArchiveZip

ForeverWintr:
If an ArchiveZip instance fails to instantiate because the filepath it was given doesn't exist, `__del__` raises an `AttributeError` because `_archive` doesn't exist.  The exception is recorded as [unraisable](https://docs.python.org/3/c-api/exceptions.html#c.PyErr_WriteUnraisable) (because it happens in `__del__`?) and is printed to stderr. This adds a test to demonstrate the issue, and fixes it.


flexatone:
Many thanks for isolating and correcting this issue.

PR #550, opened 2022-08-02
Integrate AK function.

chaburkland:


flexatone:
@chaburkland : How is the performance looking with this change? If this is ready I will merge.

chaburkland:
There isn't a single perf test that actually executes that branch!

PR #572, opened 2022-08-17
Fix saving empty frame to npz; add test

ForeverWintr:
This fixes writing an empty frame to npz. 

I also fixed reading from npz, but I'm less sure if my solution is correct there!


flexatone:
Thanks for this PR; these changes looks good. Is this ready for merging?

flexatone (approved):
Thanks for this improvement!

flexatone:
Note: codecov sometimes gets confused; this is one of those cases, I think.

PR #576, opened 2022-08-24
540/from index items support ih

### LINKED ISSUE #540: `IndexHierarchy.from_index_items` might work with pairs of label, `IndexHierarchy`
Presently `IndexHierarchy.from_index_items` fails if the pairs are pairs of label, `IndexHierarchy` (instead of pairs of label, `Index`). Conceptually, this seems like it should work. This gets called with using `Frame.from_concat_items()` with `Frame` that have `IndexHierarchy`.

@chaburkland : would you like to work on this one? 

```python
>>> ih1 = sf.IndexHierarchy.from_labels((('x', 0), ('y', 1)))
>>> ih1
<IndexHierarchy>
x                0
y                1
<<U1>            <int64>
>>> ih2 = sf.IndexHierarchy.from_labels((('p', 0), ('q', 1)))
>>> ih2
<IndexHierarchy>
p                0
q                1
<<U1>            <int64>
>>> sf.IndexHierarchy.from_index_items((('a', ih1), ('b', ih2)))
AttributeError: 'numpy.ndarray' object has no attribute '_indices'
```

chaburkland:
Closes #540


static_frame/core/index_hierarchy.py
```diff
@@ -662,6 +650,91 @@ def _repeat(i_repeat_tuple: tp.Tuple[int, int]) -> np.ndarray:
...
 
+    @classmethod
+    def from_index_items(cls: tp.Type[IH],
+            items: tp.Iterable[tp.Tuple[tp.Hashable, IndexBase]],
+            *,
+            index_constructor: IndexConstructor = None,
+            name: NameType = None,
+            ) -> IH:
+        '''
+        Given an iterable of pairs of label, :obj:`IndexBase`, produce an :obj:`IndexHierarchy` where the labels are depth 0, the indices are depth 1.
```
flexatone, line 661:
```suggestion
        Given an iterable of pairs of label, :obj:`IndexBase`, produce an :obj:`IndexHierarchy` where the labels are depth 0, the indices are depth 1. While the provided :obj:`IndexBase` can be `Index` or `IndexHierarchy`, across all pairs all depths must be the same.
```

static_frame/core/index_hierarchy.py
```diff
@@ -662,6 +650,91 @@ def _repeat(i_repeat_tuple: tp.Tuple[int, int]) -> np.ndarray:
...
+                if constructor != existing_index_constructors[i]:
+                    existing_index_constructors[i] = cls._INDEX_CONSTRUCTOR
+
+
+        outer_level = np.hstack(
+                [
+                    np.repeat(val, repeats=reps)
+                    for val, reps in zip(labels, repeats)
+                ])
+
```
flexatone, line 708:
```suggestion
        outer_level = np.hstack(
                [
                    np.repeat(val, repeats=reps)
                    for val, reps in zip(labels, repeats)
                ])
        outer_level.flags.writeable = False
```

static_frame/core/index_hierarchy.py
```diff
@@ -662,6 +650,91 @@ def _repeat(i_repeat_tuple: tp.Tuple[int, int]) -> np.ndarray:
...
+            yield outer_level
+            yield from TypeBlocks.vstack_blocks_to_blocks(blocks)
+
+        tb = TypeBlocks.from_blocks(gen_blocks())
+
+        size, depth = tb.shape
+
+        def gen_columns() -> tp.Iterator[np.ndarray]:
+            for i in range(depth):
+                yield tb._extract_array_column(i).reshape(size)
```
flexatone, line 721: Why do we need this `reshape`?

flexatone (changes requested):
Very nice work with this. Just a few questions and suggestions.

static_frame/core/index_hierarchy.py
```diff
@@ -662,6 +650,91 @@ def _repeat(i_repeat_tuple: tp.Tuple[int, int]) -> np.ndarray:
...
+                index_constructors=(index_constructor, *existing_index_constructors),
+                depth=depth,
+                )
+
+        indices, indexers = construct_indices_and_indexers_from_column_arrays(
+                column_iter=gen_columns(),
+                index_constructors_iter=index_constructors_iter,
+                )
+
+        return cls(indices=indices, indexers=indexers, name=name)
```
flexatone, line 736:
```suggestion
        return cls(indices=indices, indexers=indexers, name=name, blocks=tb)
```

static_frame/core/index_hierarchy.py
```diff
@@ -662,6 +650,91 @@ def _repeat(i_repeat_tuple: tp.Tuple[int, int]) -> np.ndarray:
...
+            for i, constructor in enumerate(index._index_constructors):  # type: ignore
+                if constructor != existing_index_constructors[i]:
+                    existing_index_constructors[i] = cls._INDEX_CONSTRUCTOR
+
+
+        outer_level = np.hstack(
+                [
+                    np.repeat(val, repeats=reps)
+                    for val, reps in zip(labels, repeats)
+                ])
```
flexatone, line 707: For most common cases this will work fine, but in some cases we will hit NumPy's undesirable type evaluation procedures, like this:
```
>>> np.hstack([np.repeat(v, repeats=r) for v, r in zip([1, 'b'], [3, 4])])
array(['1', '1', '1', 'b', 'b', 'b', 'b'], dtype='<U21')
```
I think there is the no alternative but to use `iterable_to_array_1d()` for the final array creation... or evaluate the types per element first and resolve them with `resolve_dtype`, then fill the value into a pre-allocated array.

static_frame/core/index_hierarchy.py
```diff
@@ -662,6 +650,91 @@ def _repeat(i_repeat_tuple: tp.Tuple[int, int]) -> np.ndarray:
...
+                index_constructors=(index_constructor, *existing_index_constructors),
+                depth=depth,
+                )
+
+        indices, indexers = construct_indices_and_indexers_from_column_arrays(
+                column_iter=gen_columns(),
+                index_constructors_iter=index_constructors_iter,
+                )
+
+        return cls(indices=indices, indexers=indexers, name=name, blocks=tb)
```
flexatone, line 736:
```suggestion
        return cls(indices=indices, indexers=indexers, name=name, blocks=tb, own_blocks=True)
```

flexatone (changes requested):
Great work with this, just one thing to consider.

PR #578, opened 2022-08-25
577/ih more interfaces

### LINKED ISSUE #577: Expose more IndexHierarchy interfaces
At least add:

```python
sf.IndexHierarchy.index_at_depth
sf.IndexHierarchy.indexer_at_depth
```

chaburkland:
Closes #577 

A few changes

1. Adds a repr for `sf.HLoc`
2. Removes `IndexHierarchy._get_unique_labels_in_occurence_order`
3. Adds `IndexHierarchy.index_at_depth`
4. Adds `IndexHierarchy.indexer_at_depth`
5. Adds `order_by_occurrence` parameter to `.unique(...)` calls for Index objects


static_frame/core/hloc.py
```diff
@@ -22,11 +24,53 @@ class HLoc(metaclass=HLocMeta):
...
+                yield result
+
+        if not contains_slices:
+            if len(self.key) == 1:
+                return f'HLoc[{self.key[0]}]'
+
+            # self.key is a tuple, so we strip off the parentheses.
+            return f'HLoc[{str(self.key)[1:-1]}]'
+
+        return f'HLoc[{",".join(gen_nested_keys())}]'
```
flexatone, line 76: Should we do something like this?

```suggestion
        return f'<HLoc[{",".join(gen_nested_keys())}]>'
```

static_frame/core/hloc.py
```diff
@@ -22,11 +24,53 @@ class HLoc(metaclass=HLocMeta):
...
+        # This is usually very small (i.e. <10 items), so it's very cheap to
+        # pay the price of an extra loop to check for slices.
+        for key in self.key:
+            if key.__class__ is slice:
+                contains_slices = True
+                break
+
+        def gen_nested_keys() -> tp.Iterator[str]:
+            for key in self.key:
+                if not isinstance(key, slice):
```
flexatone, line 48:
```suggestion
                if key.__class__ is slice:
```

static_frame/core/index_hierarchy.py
```diff
@@ -1649,15 +1684,15 @@ def rehierarch(self: IH,
...
         '''
-        Index could be [A, B, C]
-        Indexers could be [2, 0, 0, 2, 1]
-
-        This function return [C, A, B] -- shoutout to my initials
+        Return the unique labels in the given depth in the order they appear in the hierarchy.
         '''
+        # Index could be [A, B, C]
+        # Indexers could be [2, 0, 0, 2, 1]
+        # This function return [C, A, B] -- shoutout to my initials
```
flexatone, line 1695: hahaha

static_frame/core/index_hierarchy.py
```diff
@@ -1649,15 +1684,15 @@ def rehierarch(self: IH,
             own_blocks=True,
             )
 
-    def _get_unique_labels_in_occurence_order(self: IH,
+    def get_unique_labels_in_occurence_order(self: IH,
```
flexatone, line 1687: I was going to suggest naming this `unique_at_depth` but then I noticed the signature of `unique` already includes a `depth` parameters. Do we need this new interface, or can we extend `unique()` to do what this does (if they are different).

flexatone (changes requested):
Many thanks for these additions. 

One additional thing: as this adds new interfaces, we need new examples. Please try to run the following and, where it fails (it should for new interfaces that do not have examples_, add new examples. If you need orientation on how the example-builder works, let me know. 

`python3 doc/build_example.py`

static_frame/core/index_hierarchy.py
flexatone: Maybe we need to add  a Boolean `order_by_occurrence` flag to `unique()` on this and `Index`; that way users can opt-in to getting occurrence order.

static_frame/core/hloc.py
```diff
@@ -1,12 +1,14 @@
 import typing as tp
 
 from static_frame.core.util import GetItemKeyType
+from static_frame.core.util import NULL_SLICE
+from static_frame.core.util import EMPTY_TUPLE
```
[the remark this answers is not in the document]
chaburkland: Yeah I added it back

static_frame/core/hloc.py
```diff
@@ -22,11 +24,53 @@ class HLoc(metaclass=HLocMeta):
...
+                    result += f':{key.step}'
+
+                yield result
+
+        if not contains_slices:
+            if len(self.key) == 1:
+                return f'HLoc[{self.key[0]}]'
+
+            # self.key is a tuple, so we strip off the parentheses.
+            return f'HLoc[{str(self.key)[1:-1]}]'
```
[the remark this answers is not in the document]
chaburkland: Hah! Thanks for this

static_frame/core/hloc.py
chaburkland: I'm not sure I like this actually. It feels like it makes the repr unnecessarily artificial. Can you help me understand why you want this?

static_frame/core/index_hierarchy.py
chaburkland: I really, really like this idea!! I never liked the verbosity of `get_unique_labels_in_occurence_order`, and yet I wasn't able to come up with any ideas.

static_frame/core/hloc.py
flexatone: I do not have a strong opinion, but it is common to put class names in brackets, e.g.:

```
>>> sf.Series((1,2))                                                                                                                                                                                                                                                           
<Series>
<Index>
0        1
1        2
<int64>  <int64>
```
Unless you expect your repr to actually be "evalable"... but we do not have a lot of examples of  reprs for non-containers in StaticFrame.

static_frame/core/hloc.py
```diff
@@ -22,11 +24,37 @@ class HLoc(metaclass=HLocMeta):
...
+                    continue
+
+                if key == NULL_SLICE:
+                    yield ':'
+                    continue
+
+                if key.start is None:  # type: ignore [union-attr]
+                    result = ':'
+                else:
+                    result = f'{key.start}:' # type: ignore [union-attr]
```
flexatone, line 50: More compact and only one type ignore!
```suggestion
                result = ':' if key.start is None else f'{key.start}' # type: ignore [union-attr]
```

static_frame/core/index.py
```diff
@@ -1107,13 +1107,15 @@ def __contains__(self, value: tp.Any) -> bool:
     # utility functions
 
     def unique(self,
-            depth_level: DepthLevelSpecifier = 0
+            depth_level: DepthLevelSpecifier = 0,
+            order_by_occurrence: bool = False,
```
flexatone, line 1111: Should the default be `True`?

static_frame/core/index_hierarchy.py
```diff
@@ -1399,6 +1399,41 @@ def values_at_depth(self: IH,
...
+        Args:
+            depth_level: a single depth level, or iterable depth of depth levels.
+        '''
+        if self._recache:
+            self._update_array_cache()
+
+        if isinstance(depth_level, INT_TYPES):
+            return self._indices[depth_level]
+
+        return list(map(self._indices.__getitem__, depth_level))
```
flexatone, line 1417: Any specific reason to return a list over a tuple?
```suggestion
        return tuple(map(self._indices.__getitem__, depth_level))
```

static_frame/core/index_hierarchy.py
```diff
@@ -2165,12 +2185,14 @@ def __contains__(self: IH, # type: ignore
 
     def unique(self: IH,
             depth_level: DepthLevelSpecifier = 0,
+            order_by_occurrence: bool = False,
```
flexatone, line 2188: Same here: default to True?

flexatone (changes requested):
Very nice a work; just a few small items to consider and we can get this merged!

static_frame/core/index.py
chaburkland: I guess I was aiming for backwards compatibility here, so I wanted to retain the earlier behavior. Would you like this done by default now?

static_frame/core/index_hierarchy.py
chaburkland: Same response here: https://github.com/InvestmentSystems/static-frame/pull/578#discussion_r971392814

flexatone: I see your point. I was thinking that, for backwards compatibility with what `Index` is already doing, `True` was necessary; i.e., since `Index` was already giving the values in `order_of_occurrence`, the default there and for `IndexHierarchy` should be `True`. I know that `Index` will return the same array regardless of the value of that parameter. But I guess before these changes, `IndexHierarchy` was not returning the values in order of occurrence, right?

chaburkland: Correct. `IndexHierarchy.unique` was _not_ returning the unique values in occurrence order (unless the data was an unsortable object numpy array).

Technically,, `index.unique(order_of_occurrence=True)` == `index.unique(order_of_occurrence=False)`, since all `.unique(...)` does is return `.values`, so leaving that default to `False` will not change anything. I interpret the `order_of_occurrence=False` to mean, "We do not guarantee the values will be ordered by how they appear." It doesn't mean, "They definitely won't be ordered by how they appear."

flexatone: Ok, agreed; lets keep as you have it!

flexatone (approved):

flexatone:
@chaburkland : I will merge this in after CI completes: thank you for these contributions!

PR #579, opened 2022-08-27
Removing archive file paths on `ErrorNPYEncode`

flexatone:


flexatone:
TODO: Check if the changes in place are sufficient when using `Bus` or `Batch` to write to a zip NPZ.

PR #583, opened 2022-08-30
`Index.equals` optimization for `Index` with no `_map`

flexatone:


flexatone:
```python
In [1]: idx1 = sf.Index(np.arange(100_000), loc_is_iloc=True)                                                                                                                                           

In [2]: idx2 = sf.Index(np.arange(100_000), loc_is_iloc=True)                                                                                                                                           

In [3]: idx3 = sf.Index(np.arange(100_000), loc_is_iloc=False)                                                                                                                                          

In [4]: %timeit idx1.equals(idx2)                                                                                                                                                                       
819 ns ± 12.3 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)

In [5]: %timeit idx1.equals(idx3)                                                                                                                                                                       
73.9 µs ± 1.01 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
```

Issue #52, opened 2019-05-31
Static Typing


brandtbucher:
I nominate `mypy`, since it's another project I'm becoming more involved with. I believe this will prove useful as StaticFrame begins interfacing with C code internally. Ideally, we'd use it as a CI job too.

The following modules pass `mypy --strict`:

- [x] `doc/doc_build.py`
- [x] `doc/source/conf.py`
- [x] `setup.py`
- [x] `static_frame/__init__.py`
- [x] `static_frame/__main__.py`
- [x] `static_frame/core/__init__.py`
- [x] `static_frame/core/array_go.py`
- [x] `static_frame/core/display.py`
- [x] `static_frame/core/display_color.py`
- [x] `static_frame/core/display_html_datatables.py`
- [x] `static_frame/core/doc_str.py`
- [x] `static_frame/core/frame.py`
- [x] `static_frame/core/hloc.py`
- [x] `static_frame/core/index.py`
- [x] `static_frame/core/index_base.py`
- [x] `static_frame/core/index_hierarchy.py`
- [x] `static_frame/core/index_level.py`
- [x] `static_frame/core/iter_node.py`
- [x] `static_frame/core/operator_delegate.py`
- [x] `static_frame/core/series.py`
- [x] `static_frame/core/type_blocks.py`
- [x] `static_frame/core/util.py`
- [x] `static_frame/performance/__init__.py`
- [x] `static_frame/performance/core.py`
- [x] `static_frame/performance/main.py`
- [x] `static_frame/performance/perf_test.py`
- [x] `static_frame/performance/pydata_2018.py`
- [x] `static_frame/test/__init__.py`
- [x] `static_frame/test/property/__init__.py`
- [x] `static_frame/test/property/strategies.py`
- [x] `static_frame/test/property/test_index.py`
- [x] `static_frame/test/property/test_strategies.py`
- [x] `static_frame/test/property/test_type_blocks.py`
- [x] `static_frame/test/property/test_util.py`
- [x] `static_frame/test/test_case.py`
- [x] `static_frame/test/unit/__init__.py`
- [x] `static_frame/test/unit/test_array_go.py`
- [x] `static_frame/test/unit/test_display.py`
- [x] `static_frame/test/unit/test_display_color.py`
- [x] `static_frame/test/unit/test_doc.py`
- [x] `static_frame/test/unit/test_frame.py`
- [x] `static_frame/test/unit/test_index.py`
[...]

PR #595, opened 2022-09-07
Complete typing of `frame.py`

flexatone:


flexatone:
https://github.com/InvestmentSystems/static-frame/issues/52

Issue #182, opened 2020-03-04
Implement __sizeof__ for all container types


brandtbucher:
https://docs.python.org/3/library/sys.html#sys.getsizeof

PR #601, opened 2022-09-09
Implement getsizeof_total for Container Types

michaeljpeters:
This allows sys.getsizeof to calculate the memory usage of static frame types. #182


static_frame/core/frame.py
```diff
@@ -3006,6 +3007,18 @@ def __deepcopy__(self, memo: tp.Dict[int, tp.Any]) -> 'Frame':
     #     '''
     #     return self.__copy__() #type: ignore
 
+    def __sizeof__(self: 'Frame') -> int:
+        # TODO: Check if we need to count weakref
+        # TODO: Make sure these don't have duplicates
+        # TODO: Consider skipping counting arrays used in multiple nested objects
+        return sum(getsizeof(o) for o in [
+            self.__weakref__,
```
chaburkland, line 3015: I'm not sure we need to get the size of `__weakref__`. See: https://stackoverflow.com/a/36788452/7615851

What are your thoughts?

static_frame/core/index.py
```diff
@@ -382,6 +383,16 @@ def copy(self: I) -> I:
         '''
         return self.__copy__() #type: ignore
 
+    def __sizeof__(self: I) -> I:
+        # TODO: Could labels and positions be is-equal?
```
chaburkland, line 387: Yes they can, in which case I don't think we should double count.

Also, since `self.positions` is always going to be a slice of a global numpy array, I'm not sure `sys.getsizeof` will be accurate. Perhaps we need to use `self.positions.nbytes` instead?

static_frame/core/index.py
```diff
@@ -382,6 +383,16 @@ def copy(self: I) -> I:
...
+        return sum(getsizeof(o) for o in (
```
chaburkland, line 388: It might be a good idea to add a test that validates against the `__slots__` attribute, to guard against someone adding something to the class and forgetting to update the `__sizeof__` method.

My thought would be something like:

```python
def alternative_get_size(index: Index) -> int:
    return sum(getsizeof(getattr(index, attr) for attr in index.__slots__))

assert alternative_get_size(index) == getsizeof(index)
```

Same would probably go for the other containers

static_frame/core/type_blocks.py
```diff
@@ -531,6 +532,17 @@ def copy(self) -> 'TypeBlocks':
         '''
         return self.__copy__()
 
+    def __sizeof__(self: 'TypeBlocks') -> int:
+        # TODO: Make sure these don't have duplicates
+        return sum(getsizeof(o) for o in [
+            self._blocks,
+            *self._blocks,
```
chaburkland, line 539: This could benefit from a comment as this is a cool move :)

static_frame/core/type_blocks.py
```diff
@@ -531,6 +532,17 @@ def copy(self) -> 'TypeBlocks':
...
+            self._dtypes,
+            self._index,
+            self._shape,
+            self._row_dtype
+        ])
```
chaburkland, line 544: Also, pro-tip, this could be even slicker by using `map`.
```suggestion
        return sum(map(getsizeof, (
            self._blocks,
            *self._blocks,
            self._dtypes,
            self._index,
            self._shape,
            self._row_dtype
        ))
```
Same for the other methods

chaburkland (changes requested):
Thanks for working on this Michael! I have a few suggestions and questions.

I think it might be a good time to start adding some unit tests for these methods, and to work on resolving your TODO comments :)

static_frame/core/index.py
michaeljpeters: Working on a function that will prevent double-counting objects with the same id.

`sys.getsizeof` will return the size of the reference numpy array. Since the index is just storing a reference, I think that only the size of the reference should be counted toward the index's size.

To get the size of the global numpy array, maybe we should allow for a call to getsizeof on the PositionsAllocator?

static_frame/core/index.py
michaeljpeters: That's a beautiful test idea!

static_frame/core/frame.py
michaeljpeters: Reading that and thinking more into it, I'm against counting the __weakref__ since it's not related to the memory needed for the object to be fully functional.

static_frame/core/index.py
flexatone: @chaburkland : that will not work, as `__slots__` only defines attributes at *that* level in the object hierarchy: it does not tell you all attributes that are available (slotted) on that class (via its parents' slots).

chaburkland: Good point. I think you could use the pattern of defining module-level tuple for the slots (like you do in `index.py::_INDEX_SLOTS`) that `__slots__` is set to. Then you can test against that instead of `obj.__slots__`

michaeljpeters: Could that tuple be generated automatically? I'd rather not add an extra place to update, particularly if the purpose is to catch when people forget to update a different section of code

chaburkland: I think I wasn't very clear. My suggestion is this (you can look at `index.py` to see the existing precedent in the code base).

```python
_FRAME_SLOTS = (
            '__weakref__',
            '_blocks',
            '_columns',
            '_index',
            '_name',
            )

class Frame(ContainerOperand):
    __slots__ = _FRAME_SLOTS
```

Now, the test can just look like this:
```python
# test_frame.py
from static_frame.core.frame import _FRAME_SLOTS

def alternative_get_size(frame: Frame) -> int:
    return sum(getsizeof(getattr(index, attr) for attr in _FRAME_SLOTS))

assert alternative_get_size(frame) == getsizeof(frame)
```

static_frame/core/type_blocks.py
michaeljpeters: I like to lean toward comprehensions personally. Either way, I changed this to a function and unfortunately had to get rid of the comprehension to check for duplicated ids.

Do we generally try to use tuples instead of lists when the intention is just to create an iterable?

chaburkland: I only create lists if I need the ability to grow. In this case, we just building up a temporary container to feed its values into a function repeatedly, and so the best choice is a `tuple`, plus, `tuple` is a bit smaller than a `list`.

If you find the `map` syntax ugly, you don't need to change it, but I would recommend using a `tuple` instead of `list`. But just for fun, calling `map` is faster in this case than using a comprehension.

Performance:
```python
14:38:48  In [8] %timeit sum(sys.getsizeof(o) for o in l)
2.3 ms ± 195 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

14:38:53  In [9] %timeit sum(map(sys.getsizeof, l))
1.65 ms ± 66.3 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

Memory:
```python
14:39:49  In [10] len(l)
14:40:32 Out [10] 10000

14:39:44  In [11] sys.getsizeof(t)
14:39:47 Out [11] 80040

14:39:47  In [12] sys.getsizeof(l)
14:39:49 Out [12] 87616
```

These are all pretty small optimizations, so they are not crucial, I just think it's worth pointing out the cost/benefit analysis of the different approaches :)

static_frame/test/unit/test_index.py
```diff
@@ -1648,6 +1650,50 @@ def test_index_display_a(self) -> None:
...
+        idx1 = Index(('a', 'b', 'c'))
+        idx2 = Index(('a', 'b', 'c', 'd'))
+        self.assertTrue(getsizeof(idx1) < getsizeof(idx2))
+
+    def test_sizeof_more_nested_values_adds_size(self):
+        idx1 = Index((1, 'b', (2, 3)))
+        idx2 = Index((1, 'b', (2, 3, 4, 5)))
+        self.assertTrue(getsizeof(idx1) < getsizeof(idx2))
+    
+    def test_sizeof_more_doubly_nested_values_maintains_size(self):
```
michaeljpeters, line 1687: This is how it works now. I'm considering changing total_getsizeof to count the size of nested containers.

michaeljpeters: To do this, I'm going to recreate the "handlers" concept from the [total_sizeof recipe](https://code.activestate.com/recipes/577504/) rather than checking for instances of iterables within the containers.

This way, iterables that represent the full contents in __sizeof__ (like numpy arrays) can be handled differently than those that do not (like lists of lists).

static_frame/core/util.py
```diff
@@ -3187,15 +3187,27 @@ def iloc_to_insertion_iloc(key: int, size: int) -> int:
...
-        if id(e) in seen:
+    for el in it:
+        if id(el) in seen:
             continue
-        seen.add(id(e))
-        total += getsizeof(e)
+        seen.add(id(el))
+        total += getsizeof(el)
+        for t, h in handlers.items():
+            if isinstance(el, t):
```
flexatone, line 3212: I feel like this approach is bit to heavy... we do not need to try every element in `dict`, we just need to handle the one special case: `dict`. I think better is not define `handlers`, and instead just branch here when you find a `dict`.

static_frame/test/unit/test_frame.py
```diff
@@ -14560,7 +14560,7 @@ def test_frame_bloc_sel_a(self) -> None:
     #---------------------------------------------------------------------------
     def test_sizeof_a(self) -> None:
         f = ff.parse('s(3,4)')
-        self.assertEqual(getsizeof(f), 976)
+        self.assertEqual(getsizeof(f), 1336)
```
flexatone, line 14563: These are the kinds of tests that might be very fragile and fail on Windows or different versions of Python / NumPy...

static_frame/test/unit/test_series.py
```diff
@@ -5387,9 +5387,35 @@ def test_sizeof_simple(self):
...
+    def test_sizeof_with_index(self):
+        # This is notably larger since the index is not loc_is_iloc optimised
         s = Series(('a', 'b', 'c'), index=(0, 1, 2))
-        self.assertEqual(getsizeof(s), 292)
+        self.assertEqual(getsizeof(s), 932)
+
+    def test_sizeof_with_name_is_larger(self):
+        s1 = Series(('a', 'b', 'c'))
+        s2 = s1.rename('named_series')
+        self.assertTrue(getsizeof(s1) < getsizeof(s2))
```
flexatone, line 5398: This is an interesting approach that is robust to platform changes!

static_frame/test/unit/test_frame.py
flexatone: e.g.,: https://github.com/InvestmentSystems/static-frame/actions/runs/3038997283/jobs/4893403939

static_frame/core/util.py
michaeljpeters: While `dict` is the special case, elements of types not included handlers must not be looped on (e.g. `int`, `str`, `FrozenAutoMap`). So we need a branch here based on the element's type being in the handlers and then special handling based on if the element is a dict/ndarray.

static_frame/test/unit/test_frame.py
michaeljpeters: I will look into a better way of making these work on different operating systems/versions of python.

static_frame/core/index.py
```diff
@@ -382,6 +383,15 @@ def copy(self: I) -> I:
         '''
         return self.__copy__() #type: ignore
 
+    def __sizeof__(self: I) -> int:
```
michaeljpeters, line 386: Something to consider:
`__sizeof__` only counts the size of the container objects and not its contents for things like list.

E.g. `[2, 3, 4]` has the same size as `['abc', 'def', 'ghi']` since they're both lists with 3 elements.

Maybe we should follow this paradigm with StaticFrame containers too. If we do this, to count the full size of the container and its contents, we could add handlers in the total_getsizeof utility function..

static_frame/core/util.py
flexatone: Ah, I see... maybe you just need to instead look for non-string iterable objects, i.e.: `hasattr(obj, '__len__') and not isinstance(obj, str)`. Calling `isinstance` many times for every object might get costly.

static_frame/core/index.py
flexatone: I think it is better to try to show true inclusive memory foot print, as that is what I think most users want (and that is pretty much what NumPy is doing). I am not sure why common containers give up at the first level... maybe for performance... but it just does not seem useful to only show one depth level of size.

static_frame/core/util.py
```diff
@@ -3186,5 +3187,27 @@ def iloc_to_insertion_iloc(key: int, size: int) -> int:
...
+    Gives the total size of an iterable of elements
+    see also: https://code.activestate.com/recipes/577504/
+    '''
+    handlers = {
+        tuple: iter,
+        list: iter,
+        dict: lambda d: chain.from_iterable(d.items()),
+        set: iter,
+        frozenset: iter,
+        np.ndarray: lambda d: iter(d) if d.dtype == np.object else []
```
flexatone, line 3201:
```suggestion
        np.ndarray: lambda d: iter(d) if d.dtype == np.object else ()
```

static_frame/core/util.py
```diff
@@ -3186,5 +3187,27 @@ def iloc_to_insertion_iloc(key: int, size: int) -> int:
         raise IndexError(f'index {key} out of range for length {size} container.')
     return key % size
 
-
-
+def total_getsizeof(it: tp.Iterable[any], *, seen=None) -> int:
```
flexatone, line 3190: We try to use "hierarchical naming", meaning you name from the general to more specific, which may not always be obvious, but can be really helpful for consistency and intuitions about what things are named. I would consider naming this function `'getsizeof_total()` or `getsizeof_recursive()`, putting the `getsizeof` bit first.

static_frame/test/unit/test_type_blocks.py
```diff
@@ -4030,6 +4031,78 @@ def test_unified_dtypes_c(self) -> None:
...
+    def test_sizeof_b(self) -> None:
+        tb = TypeBlocks.from_blocks([
+            np.array([1, 2, 3]),
+            np.array([4, 5, 6])
+        ])
+        self.assertEqual(getsizeof(tb), sum(getsizeof(e) for e in [
+            # _blocks
+            np.array([1, 2, 3]),
+            np.array([4, 5, 6]),
+            tb._blocks, # [np.array([1, 2, 3]), np.array([4, 5, 6])],
```
michaeljpeters, line 4066: The `getsizeof` size of `_blocks` is not equal to the size of the constructed list. from_blocks uses append to create the _blocks array. Using append triggers allocation for a few more elements. I used references to the member variables rather than constructing these manually to mitigate this inconsistency.
```
>>> a = []
>>> a.append(2)
>>> b = [2]
>>> getsizeof(a)
88
>>> getsizeof(b)
64
```

static_frame/core/type_blocks.py
michaeljpeters: I'm on-board with using a tuple in cases where things don't need to grow. Interesting to see the performance benefit of map. I am still not a big fan of the map syntax so I plan to leave in the comprehensions.

static_frame/test/unit/test_index.py
michaeljpeters: Doubly-nested values no longer maintain size.

static_frame/core/util.py
michaeljpeters: Unfortunately there are a couple iterable objects that we don't want to loop over (like `FrozenAutoMap`). I think running a `hasattr(obj, '__len__')` could be a good thing to check before going to a bunch of `isinstance` calls if we are worried about performance cost.

static_frame/core/util.py
michaeljpeters: Renamed it to `getsizeof_recursive` to follow this convention.

static_frame/core/index.py
michaeljpeters: Got double-count prevention working!

static_frame/core/util.py
flexatone: Interesting... why do you not want to count the keys in a `FrozenAutoMap`?

static_frame/test/unit/test_type_blocks.py
flexatone: That is an interesting observation. I think our goal here is to find memory foot print of SF containers. If, for example, we want to multiprocess a large `Frame` we would need to know how much memory we might need in each process. That would include over-allocation from mutable containers. So maybe it is okay (better?) to included those over-allocated positions. You could also make it an option into your utility function.

michaeljpeters: I think it's better to include the over-allocated positions in the default results.

Are you suggesting to include an option to compute the "minimal allocation" for mutable containers like lists in the utility function? I am not sure that such a statistic would be useful unless we are looking for places to optimize (such as replacing mutable containers with minimally allocated ones) or launching a different thread somehow uses the "minimal allocation"

flexatone: Great, I agree: lets count the over-allocated positions. I do not see a strong need to implement a "minimal allocation" option: I just thought you were implying it was your preference!  But no problem only doing it one way.

static_frame/core/util.py
```diff
@@ -3186,5 +3187,34 @@ def iloc_to_insertion_iloc(key: int, size: int) -> int:
...
+    total = 0
+    seen = set() if seen is None else seen
+    if id(obj) in seen:
+        return 0
+    seen.add(id(obj))
+    total += getsizeof(obj)
+    # Check if iterable or a string first for fewer isinstance calls on common types
+    if hasattr(obj, '__iter__') and not isinstance(obj, str):
+        if (
+            (isinstance(obj, np.ndarray) and obj.dtype == np.object)
```
michaeljpeters, line 3207: Swap `obj.dtype == np.object` with `obj.dtype is np.object`?

flexatone: Common in SF and fastest is the following. Note that we cannot do `is` comparisons on dtypes as there are many instances of object dtype...

```suggestion
            (obj.__class__ is np.ndarray and obj.dtype.kind == DTYPE_OBJECT_KIND)
```

static_frame/core/index.py
michaeljpeters: Added `sizeof_helper` (final name tbd) to automatically collect the slots in each level of the hierarchy. I think the test cases should check that all class attributes are accounted for manually.

static_frame/core/util.py
```diff
@@ -3186,5 +3187,43 @@ def iloc_to_insertion_iloc(key: int, size: int) -> int:
...
+    seen.add(id(obj))
+    for el in get_unsized_children_iter(obj):
+        yield from all_nested_elements(el, seen=seen)
+    yield obj
+
+def getsizeof_recursive(obj: any, *, seen=None) -> int:
+    seen = set() if seen is None else seen
+    return sum(getsizeof(el) for el in all_nested_elements(obj, seen=seen))
+
+def sizeof_helper(cls, self, *, seen):
```
michaeljpeters, line 3221: Wanted to just sum each slot element with something like this but I'm not sure it's possible for python to know the MRO.
```
def collect_slots(cls, self):
    if hasattr(cls, '__slots__'):
        yield from collect_slots(super(cls, self), self)
        yield from cls.__slots__
```

michaeljpeters:
```python
def collect_slots(obj):
    return frozenset().union(cls.__slots__ for cls in obj.__class__.__mro__)
```

static_frame/core/util.py
```diff
@@ -3186,5 +3187,38 @@ def iloc_to_insertion_iloc(key: int, size: int) -> int:
...
+    if id(obj) in seen:
+        return
+    seen.add(id(obj))
+    for el in _unsized_children(obj):
+        yield from _nested_sizable_elements(el, seen=seen)
+    yield obj
+
+def getsizeof_recursive(obj: any, *, seen=None) -> int:
+    seen = set() if seen is None else seen
+    return sum(getsizeof(el) for el in _nested_sizable_elements(obj, seen=seen))
```
flexatone, line 3219: This is the kind of collection of functions you can put in a class-only Class that groups these functions (as static and class methods) together under a common namespace.

static_frame/core/util.py
```diff
@@ -3186,5 +3187,38 @@ def iloc_to_insertion_iloc(key: int, size: int) -> int:
...
 
+def _unsized_children(obj):
+    # Check if iterable or a string first for fewer isinstance calls on common types
+    if hasattr(obj, '__iter__') and not isinstance(obj, str):
+        if (
+            (obj.__class__ is np.ndarray and obj.dtype.kind == DTYPE_OBJECT_KIND)
+            or isinstance(obj, tuple)
+            or isinstance(obj, list)
+            or isinstance(obj, set)
+            or isinstance(obj, frozenset)
```
flexatone, line 3198: This can use `abc.Set`:
```
>>> from collections import abc                                                 
>>> isinstance(set(), abc.Set)                                                  
True
>>> isinstance(frozenset(), abc.Set)                                            
True
```

static_frame/core/util.py
```diff
@@ -3186,5 +3187,38 @@ def iloc_to_insertion_iloc(key: int, size: int) -> int:
         raise IndexError(f'index {key} out of range for length {size} container.')
     return key % size
 
+def _unsized_children(obj):
+    # Check if iterable or a string first for fewer isinstance calls on common types
+    if hasattr(obj, '__iter__') and not isinstance(obj, str):
+        if (
+            (obj.__class__ is np.ndarray and obj.dtype.kind == DTYPE_OBJECT_KIND)
+            or isinstance(obj, tuple)
+            or isinstance(obj, list)
```
flexatone, line 3196: This can use `abc.Sequence`:
```
>>> isinstance([], abc.Sequence)                                                
True
>>> isinstance((), abc.Sequence)                                                
True
```

michaeljpeters:
Useful discussion on the purpose for sys.getsizeof. https://github.com/python/cpython/issues/63248 . It has an interesting gettotalsizeof function that I didn't see in other places.
Going to look into using `gc.get_referants` instead of the "collect_slots" implementation. Also I think we should consider making this a different function from `__sizeof__` since we would otherwise be diverging from the standard (i.e. not counting references)

static_frame/test/property/test_getsizeof_total.py
```diff
@@ -0,0 +1,102 @@
...
+        >>>
+        >>> id(a[0])
+        140280362690400
+        >>> id(a[0])
+        140280363921536
+        >>> id(a[0])
+        140280363921536
+        >>> id(a[0])
+        140280363921536
+        '''
```
michaeljpeters, line 78: This is failing because this form of numpy object array doesn't always return an object with the same ID. When this fails, it's off by 96, the size of one of the numpy arrays. Annoyingly, it does not fail consistently. For the most part (all the unit tests at least), this is not a problem. The property tests that make this fail have a numpy object array that contains multiple numpy object arrays of None (as shown in line 60).

Need to find out if there is a consistent way to get the contents of a numpy array for size counting. Alternatively, maybe the inconsistency is actually accurate to the memory that numpy is using?

flexatone: That is very surprising!

ForeverWintr: I'm not sure if this is relevant here, as the dtype of `a` is object, but this answer suggests that because numpy does not store its array elements as python objects, it may create a new 'wrapper' python object on each access. I think this explanation is plausible if numpy uses a special placeholder for `None`. 

https://stackoverflow.com/a/39355706/1286571

michaeljpeters: I'm removing a comment I had in code but removed. It may be useful to understand what I did to alleviate this problem.

```
        # The getsizeof_total function does not consistently size numpy object arrays since
        # numpy object arrays of numpy object arrays do not always return consistent elements.
        # While the ids of each element of a are the same, they may appear different during iteration.
        # For the purposes of only counting each unique element once, 
        #elements with unique ids, we cannot loop over each element in the numpy array to collect a list of unique objects
        '''
        >>> import numpy as np
        >>> a = np.array([np.array([None, None, None]) for i in range(5)])
        >>>
        >>> id(a[0]), id(a[1]), id(a[2]), id(a[3]), id(a[4])
        (140036333154352, 140036333154352, 140036333154352, 140036333154352, 140036333154352)
        >>> [id(n) for n in a]
        [140036333154352, 140036331924176, 140036333154352, 140036331924176, 140036333154352]
        '''
        '''
        >>> import numpy as np
        >>> a = np.array([np.array([None, None, i]) for i in range(5)])
        >>> b = [el for el in a]
        >>> c = [id(el) for el in a]
        >>>
        >>> a
        array([[None, None, 0],
               [None, None, 1],
               [None, None, 2],
               [None, None, 3],
               [None, None, 4]], dtype=object)
        >>> b
        [array([None, None, 0], dtype=object), array([None, None, 1], dtype=object), array([None, None, 2], dtype=object), array([None, None, 3], dtype=object), array([None, None, 4], dtype=object)]
        >>> c
        [139668158326624, 139668158326544, 139668158326624, 139668158326544, 139668158326624]
        >>> set(c)
        {139668158326624, 139668158326544}
        '''
```

static_frame/core/util.py
michaeljpeters: Moved the helper functions to `MemoryMeasurements` but kept `getsizeof_total` outside of the class

static_frame/test/property/test_getsizeof_total.py
michaeljpeters: Figured out that by serializing each numpy object array element to a list, they will no longer have the duplicate id problem. See also: comment in `_unsized_children`.

static_frame/core/util.py
```diff
@@ -3186,5 +3187,91 @@ def iloc_to_insertion_iloc(key: int, size: int) -> int:
...
+                #        [None, None, 2]], dtype=object)
+                # >>> b
+                # [139999659995264, 139999659995264, 139999659995264, 139999659995264, 139999659995264]
+                # >>> c
+                # [139999659995264, 139999658764368, 139999659995264, 139999658764368, 139999659995264]
+                # >>> d
+                # [array([None, None, 0], dtype=object), array([None, None, 0], dtype=object), array([None, None, 1], dtype=object), array([None, None, 1], dtype=object), array([None, None, 2], dtype=object)]
+                # >>> e
+                # [139999659995264, 139999658764368, 139999658765088, 139997430769344, 139997430769104]
+                yield from [el for el in obj]
```
flexatone, line 3227: As discussed, I think this should do the same without the list... or you could use `iter(obj)`, but as you point out, we may get unexpected results if we have an object array that contains NumPy arrays. If this covers all cases better, fine to have this intermediary list, lets just make sure it is clear to the reader why.
```suggestion
                yield from obj
```

static_frame/core/util.py
```diff
@@ -3186,5 +3187,91 @@ def iloc_to_insertion_iloc(key: int, size: int) -> int:
...
+        if id(obj) in seen:
+            return
+        seen.add(id(obj))
+
+        for el in MemoryMeasurements._unsized_children(obj):
+            yield from MemoryMeasurements._nested_sizable_elements(el, seen=seen)
+        for el in MemoryMeasurements._sizable_slot_attrs(obj):
+            yield from MemoryMeasurements._nested_sizable_elements(el, seen=seen)
+
+        yield obj
```
flexatone, line 3268: This can be `classmethod` to avoid explicit reference to `MemoryMeasurements`

```suggestion
    @classmethod
    def nested_sizable_elements(cls, obj: tp.Any, *, seen: tp.Set[int]) -> tp.Iterable[tp.Any]:
        '''
        Generates an iterable of all objects the parent object has references to, including nested references.
        This function considers both the iterable unsized children (based on _unsized_children) and the sizable
        attributes listed in its slots. The resulting generator is in pre-order and includes the parent object
        at the end.
        '''
        if id(obj) in seen:
            return
        seen.add(id(obj))

        for el in cls._unsized_children(obj):
            yield from cls._nested_sizable_elements(el, seen=seen)
        for el in cls._sizable_slot_attrs(obj):
            yield from cls._nested_sizable_elements(el, seen=seen)

        yield obj
```

static_frame/core/util.py
```diff
@@ -3186,5 +3187,91 @@ def iloc_to_insertion_iloc(key: int, size: int) -> int:
...
+
+def getsizeof_total(obj: tp.Any, *, seen: tp.Union[None, tp.Set[tp.Any]] = None) -> int:
+    '''
+    Returns the total size of the object and its references, including nested refrences
+    '''
+    seen = set() if seen is None else seen
+    total = sum(getsizeof(el) for el in MemoryMeasurements._nested_sizable_elements(obj, seen=seen))
```
flexatone, line 3275: Lets make the public entry point explicit:
```suggestion
    total = sum(getsizeof(el) for el in MemoryMeasurements.nested_sizable_elements(obj, seen=seen))
```

flexatone (changes requested):
Very nice work on this, and I appreciate the very thorough testing. Just a few small things and we can get this merged in.

static_frame/core/util.py
michaeljpeters: I just realized that numpy is actually creating a multi-dimensional object array, even though I passed in a list of np.array objects to the constructor. The following will loop just on the objects referenced in the array.

Previously, numpy was creating new numpy arrays for each column as I looped through. This is likely why the ids were not consistent with the ids of the arrays I used in the constructor.

```
>>> import numpy as np
>>> a = np.array([np.array([None, None, i // 2]) for i in range(5)])
>>>
>>> a.shape
(5, 3)
>>>
>>> [id(a[loc]) for loc in np.ndindex(a.shape)]
[94341345798752, 94341345798752, 94341345978080, 94341345798752, 94341345798752, 94341345978080, 94341345798752, 94341345798752, 94341345978112, 94341345798752, 94341345798752, 94341345978112, 94341345798752, 94341345798752, 94341345978144]
```

static_frame/test/property/test_getsizeof_total.py
michaeljpeters: Even better, looping exclusively on the object references. See util.py:3202.

flexatone (approved):

flexatone:
Thanks again, @michaeljpeters !

PR #622, opened 2022-09-22
Return booleans from via_str.contains

ForeverWintr:
I think the intent was to return bools, not strings! :)


flexatone:
Great, thanks for this fix! Please also re-render the examples with:
```
python3 doc/build_example.py --write
```

flexatone (approved):
Many thanks for this fix!

Issue #628, opened 2022-09-28
Improve IH Error Message When Given Non-Unique Records


chaburkland:
**Example**

```python
>>> sf.Frame.from_records([['c', 'd'], ['c', 'd']], columns=['a', 'b']).set_index_hierarchy(['a', 'b'])
ErrorInitIndexNonUnique: 0
```

PR #629, opened 2022-09-28
Improves a LocMap error message

chaburkland:
Refs #628


static_frame/core/loc_map.py
```diff
@@ -41,6 +41,13 @@
 _ZERO_PAD_ARRAY.flags.writeable = False
 
 
+class FirstDuplicatePosition(KeyError):
+    __slots__ = ('first_dup',)
```
flexatone, line 45: While I appreciate the sentiment, I am not sure this `__slots__` definition unless `KeyError` and all of its parents are slotted.

PR #639, opened 2022-10-05
Start YouTube link at start of lightning talk

eli-b:
For now, I did not change the other link to a lightning talk to point directly to YouTube, which would enable a similar change.


flexatone (approved):
Thanks for this improvement!

PR #643, opened 2022-10-07
`IndexHierachy.from_values_per_depth()`

flexatone:


static_frame/core/index_hierarchy.py
```diff
@@ -447,22 +447,36 @@ def _from_arrays(cls: tp.Type[IH],
...
+                    raise ErrorInitIndex('per depth iterables must be the same length')
+
             # NOTE: we are not checking that they are all 1D
             depth = len(arrays)
             column_iter = arrays
 
         if not size:
-            return cls._from_empty((), name=name, depth_reference=depth_reference)
+            if depth is None:
+                raise RuntimeError('depth_reference must be specified for empty values')
```
chaburkland, line 478: Thanks for catching this!

static_frame/core/index_hierarchy.py
```diff
@@ -474,17 +488,24 @@ def _from_arrays(cls: tp.Type[IH],
...
 
-        blocks = TypeBlocks.from_blocks(arrays)
 
         return cls(
                 indices=indices,
                 indexers=indexers,
                 name=name,
                 blocks=blocks,
-                own_blocks=True,
+                own_blocks=own_blocks,
```
chaburkland, line 508: `own_blocks` is always `True` per the above branch

chaburkland (approved):
Awesome! Thanks for extending this

static_frame/core/index_hierarchy.py
flexatone: Good catch; it should be False on 497!

chaburkland: @flexatone Did you fix this? line 497 still says `own_blocks=True`?

flexatone: Thanks for reminding me, I am surprised it did not result in an error!

PR #651, opened 2022-10-18
Integrate `delimited_to_arrays`

flexatone:


flexatone:
Changes:

The old implementation would first rely on `genfromtxt`, and then astype types as needed. This means that all integers could be cast to bool.

PR #667, opened 2022-11-02
URL Accessor

flexatone:


static_frame/core/url.py
```diff
@@ -0,0 +1,94 @@
...
+        names = zf.namelist()
+        if len(names) > 1:
+            raise RuntimeError(f'more than one file found in zip archive: {names}')
+        print(names)
+
+def URL(url: str,
+        encoding: tp.Optional[str] = 'utf-8',
+        in_memory: bool = True,
+        buffer_size: int = 8192,
+        ) -> tp.Union[Path, StringIO, BytesIO]:
```
gettodaze, line 80: I'm not sure if this is a common use case, but it might be nice to be able to provide request headers here. I see [urlopen](https://docs.python.org/3/library/urllib.request.html#urllib.request.urlopen) allows the url to be a `urllib.request.Request` object which takes in headers. Alternatively maybe allowing the url to be a `Request` object could give that flexibility.

flexatone: Great point! I have adjusted the type hints and code to handle both an `str` or a `Request` object.

static_frame/core/url.py
```diff
@@ -0,0 +1,318 @@
...
+            extension: str,
+            ) -> tp.Union[Path, BytesIO]:
+        archive: tp.Union[Path, BytesIO]
+
+        with request.urlopen(url) as response:
+            if in_memory:
+                archive = BytesIO(response.read())
+            else:
+                with tempfile.NamedTemporaryFile(mode='wb',
+                        suffix='zip',
```
flexatone, line 116: Set extension passed in.

PR #708, opened 2022-12-28
lets -> let's

nathanjmcdougall:


flexatone:
@nathanjmcdougall : thanks for this correction; I will merge it in shortly.

PR #723, opened 2023-01-24
`HLoc` selection in `IH`

flexatone:


flexatone:
@chaburkland : bringing back the spirit of the `single_depth` optimization led me to a further optimization: getting rid of the 2D Boolean array entirely!

PR #734, opened 2023-01-31
Improving `arrays_equal`

flexatone:


static_frame/core/util.py
```diff
@@ -2005,6 +2010,15 @@ def arrays_equal(array: np.ndarray,
     '''
     Given two arrays, determine if they are equal; support skipping Na comparisons and handling dt64
     '''
+    if id(array) == id(other):
+        return True
+
+    if (mloc(array) == mloc(other)
```
chaburkland, line 2016: Is this not using `array_signature` to save on function-call overhead? I only ask cause I think the code would read more clearly if it was:
```python
if array_signature(array) == array_signature(other):
    return True
````

flexatone: Yes, minimizing function call overhead and object creation for what is I think a minority scenario. This way we pay the least for this check. I agree it would be better to use `array_signature`, but heavier, and this is a frequently called routine.

PR #744, opened 2023-02-10
Do not include `_weak_cache` in `Store` pickle state

michaeljpeters:
Removes `_weak_cache` from `Store` pickle state to allow pickleing and therefore easier use in multi processing.

https://github.com/static-frame/static-frame/issues/743


flexatone:
@michaeljpeters : just one more thing: run `inv format` to get isort to sort the imports and this should be good to go.

flexatone:
Thanks again for this contribution.

PR #767, opened 2023-03-13
Adds corrected & specialized function for quilt construction when component frames have index hierarchies on axis of selection

### LINKED ISSUE #766: Quilt iloc selection bug
**Description**

When creating a quilt from frames that have hierarchical indices, and those indices are not in tree-order, subsequent selections are inaccurate

**Example**
```python
import frame_fixtures as ff
import static_frame as sf

frame = ff.parse('f(Fg)|v(int,bool,str)|i((I,I),(str,int))|s(12,2)')
frame = frame.iloc[np.array([11, 0, 4, 9, 1, 10, 5, 2, 6, 3, 7, 8])]

f1 = frame.iloc[:6].rename("f1")
f2 = frame.iloc[6:].rename("f2")

bus = sf.Bus.from_frames((f1, f2))
quilt = sf.Quilt(bus, retain_labels=False, axis=0)

selected_index = quilt.loc[sf.HLoc["ztsv"]].index

# This fails!
assert set(selected_index.values_at_depth(0)) == {"ztsv"}
```

**Platform**
```python
<Series: platform>
<Index>
platform           Linux-5.4.0-144-generic-x86_64-with-glibc2.29
sys.version        3.8.10 (default, Nov 14 2022, 12:59:47) [GCC 9.4.0]
static-frame       1.2.7
numpy              1.19.5
pandas             1.3.5
xlsxwriter         1.1.2
openpyxl           3.0.9
xarray             0.13.0
tables             3.7.0
pyarrow            6.0.1
msgpack            (1, 0, 0)
msgpack_numpy      None
<<U13>             <object>

```

chaburkland:
Fixes #766


static_frame/core/axis_map.py
```diff
@@ -32,6 +34,49 @@ def get_extractor(
...
+
+    items_iter = iter(bus.items())
+
+    primary = IndexHierarchyGO(get_next(items_iter))
+
+    while True:
+        try:
+            primary.extend(get_next(items_iter))
+        except StopIteration:
+            break
```
flexatone, line 75: Maybe just:

```suggestion
    for ih in get_next(items_iter):
        primary.extend(ih)
```

static_frame/core/index_hierarchy.py
```diff
@@ -2784,16 +2780,15 @@ def level_add(self: IH,
         index_cls = self._INDEX_CONSTRUCTOR if index_constructor is None else index_constructor._MUTABLE_CONSTRUCTOR # type: ignore
```
flexatone, line 2780: This is old code, but it looks like we are always getting the `_MUTABLE_CONSTRUCTOR`; is that correct?

static_frame/core/index_hierarchy.py
```diff
@@ -2784,16 +2780,15 @@ def level_add(self: IH,
...
 
         if self.STATIC:
-            indices = [index_cls((level,)), *self._indices]
+            indices = [index_cls((level,))] + self._indices
```
flexatone, line 2783: Maybe:
```suggestion
            indices = [index_cls((level,)), *self._indices]
```

flexatone (changes requested):
Thanks for fixing this! Just a few small suggestions for you and then we can get this merged in.

static_frame/core/index_hierarchy.py
chaburkland: I'm not sure what you mean... If `index_constructor is None`, then we get `self._INDEX_CONSTRUCTOR`, else, `index_constructor._MUTABLE_CONSTRUCTOR`

Regardless, this doesn't seem right. I think it should be:
```python
index_cls = self._INDEX_CONSTRUCTOR if index_constructor is None else index_constructor
```
Thoughts?

flexatone: Well, we are trying to do a few things at once here.

On review, I can see that `_INDEX_CONSTRUCTOR` is set to either GO or non-GO depending on if the class is `IndexHierarchyGO` or `IndexHierarchy`, respectively. So we do not need to vary that parameter if we are using it. But, if we get an `index_constructor` argument, we want to try to use the right variant. So I think what we want to do is this,

Some tests around this variability will help us see if we are covering all cases.

```python
if index_constructor is None:
    index_cls = self._INDEX_CONSTRUCTOR 
else: 
    index_cls = index_constructor._IMMUTABLE_CONSTRUCTOR if self.STATIC else index_constructor._MUTABLE_CONSTRUCTOR
```

flexatone:
@chaburkland : that one Windows failures does not seem to be due to Hypothesis strangeness: any ideas?

static_frame/core/axis_map.py
```diff
@@ -33,6 +35,46 @@ def get_extractor(
...
+        return axis0.level_add(label)
+
+    items_iter = iter(bus.items())
+
+    primary = IndexHierarchyGO(get_next(items_iter))
+
+    for _ in range(1, len(bus)):
+        primary.extend(get_next(items_iter))
+
+    return IndexHierarchy(primary), opposite # type: ignore
```
flexatone, line 75: I think it is better if the function processes pairs only, and does not do the iteration. Then we get a more natural iteration at the end. Also, I think we can remove some asserts and replace plausible errors with exceptions. Finally, since this is an internally managed IHg, we can strip out its TypeBlocks for usage in our final returned IH.

```suggestion
    def level_add(tp.Tuple[label: tp.Hashable, frame=Frame]) -> IndexHierarchy:
        nonlocal opposite

        if axis == 0:
            axis0, axis1 = extractor(frame.index), frame.columns
        else:
            axis0, axis1 = extractor(frame.columns), frame.index

        if opposite is None:
            opposite = extractor(axis1)
        else:
            if not opposite.equals(axis1):
                raise init_exception_cls('opposite axis must have equivalent indices')

        if axis0._NDIM != 2:
             raise RuntimeError('bad index') # something better!
        return axis0.level_add(label)

    items_iter = iter(bus.items())

    primary = IndexHierarchyGO(level_add(next(items_iter)))

    for level in items_iter:
        primary.extend(level_add(level))

    return IndexHierarchy._from_type_blocks(primary._blocks, ), opposite # type: ignore
```

static_frame/core/index_hierarchy.py
```diff
@@ -2784,19 +2780,21 @@ def level_add(self: IH,
...
-                [
-                    np.zeros(self.__len__(), dtype=DTYPE_INT_DEFAULT),
-                    *self._indexers
-                ],
-                dtype=DTYPE_INT_DEFAULT,
+        indexers = np.vstack(
+            (
+                np.zeros(self.__len__(), dtype=DTYPE_INT_DEFAULT),
+                self._indexers,
+            )
```
flexatone, line 2797: This might be more efficient:
```suggestion
        indexers = np.zeros(len(self._indexers) + 1, self.__len__(), dtype=DTYPE_INT_DEFAULT)
        indexers[1:] = self._indexers 
```

flexatone (changes requested):
Thanks for these improvements!  Just a few small suggestions.

static_frame/core/axis_map.py
chaburkland: Good suggestion! One small note: `IndexHierarchy(primary)` is much faster than `IndexHierarchy._from_type_blocks(primary._blocks)`. This is because you have to re-create a bunch of different components inside of `_from_type_blocks`, whereas the specialized path in `__init__` can borrow all references to static internally managed components without having to do any analysis

flexatone (approved):
Looks good, thanks!

PR #811, opened 2023-06-23
Fix .loc Type Hints for Compound Containers

michaeljpeters:
Submitting a tuple to .loc on Compound containers (like Frame) performs a multi-dimensional access. These changes fix the type-hints so that loc/iloc can accept tuples.
```
>>> f
<Frame>
<Index> a       b       c       d       <<U1>
<Index>
0       2       8       4       9
1       3       9       8       2
2       4       10      2       1
<int64> <int64> <int64> <int64> <int64>
>>> f.loc[0, "a"]
2
```

### DIFF
--- a/static_frame/core/batch.py
+++ b/static_frame/core/batch.py
@@ -17,7 +17,7 @@ from static_frame.core.node_dt import InterfaceBatchDatetime
 from static_frame.core.node_fill_value import InterfaceBatchFillValue
 from static_frame.core.node_re import InterfaceBatchRe
 from static_frame.core.node_selector import InterfaceBatchAsType
-from static_frame.core.node_selector import InterfaceGetItem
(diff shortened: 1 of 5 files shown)


flexatone:
Thanks, @michaeljpeters , for this improvement!

Please see if you can get all the tests to pass by updating the code in `interface.py` that was matching explicitly on  `InterfaceGetItem`:

```python
            elif isinstance(obj, InterfaceGetItem) or name == cls.GETITEM:
                yield from InterfaceRecord.gen_from_getitem(**kwargs)
```

This needs to be updated to match both `InterfaceGetItem` and `InterfaceGetItemCompound` now, I think.

michaeljpeters:
All Passing! Thanks for the hint of where to modify.

PR #830, opened 2023-08-17
Containers defined with `Generic`

flexatone:
Remaining tasks:
- [x] Update project description
- [x] Determine if `FrameGO` should be generic by columnar type
- [x] Add `Require.Shape`
- [x] Update example generation to include `via_type_clinc`
- [x] Update examples to include `TypeClinic`, `ClinicResult`, `CallGuard`
- [x] Extend `Labels` validator to permit arbitrary column / row checks
- [x] Make Bus/Yarn generic
- [x] Update mypy
- [x] Update docs to include `TypeClinic`, `CallGuard`, etc.
- [x] Determine where to put / how to doc `check_interface`
- [x] Implement `TypeClinic.warn()`
- [x] Update interface displays / docs for `via_typeclinic`
- [x] Restore doc strings on `__init__`
- [x] Documentation of `consolidate` not working right, `Frame.consolidate()` is repeated twice.


flexatone:
Remaining tasks:
- [ ] Make Bus/Yarn generic
- [ ] Update interface displays for `via_typeclinic`
- [ ] Update docs to include TypeClinic
- [ ] Extend `Labels` validator to permit arbitrary column / row checks
- [ ] Determine where to put / how to doc `check_interface`
- [ ] Restore doc strings on __init__
- [ ] Update project description

chaburkland (commented):
I'll submit general feedback in chunks, so I don't fall prey to the trap of wanting to finish reviewing everything and never finishing

static_frame/core/bus.py
```diff
@@ -864,7 +869,7 @@ def unpersist(self) -> None:
     #---------------------------------------------------------------------------
     # extraction
 
-    def _extract_iloc(self, key: TILocSelector) -> 'Bus':
+    def _extract_iloc(self, key: TILocSelector) -> tp.Self:
```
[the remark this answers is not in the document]
flexatone: The right solution here is to use @overload... which you will see I use heavily in `Series` and `Frame`... I just have not yet implemented the right overload here yet. Easy to add later is my thinking.

static_frame/core/container_util.py
```diff
@@ -67,17 +68,19 @@
...
     from static_frame.core.quilt import Quilt  # pylint: disable=W0611,C0412 #pragma: no cover
     from static_frame.core.series import Series  # pylint: disable=W0611,C0412 #pragma: no cover
     from static_frame.core.type_blocks import TypeBlocks  # pylint: disable=W0611,C0412 #pragma: no cover
 
-    NDArrayAny = np.ndarray[tp.Any, tp.Any] # pylint: disable=W0611 #pragma: no cover
-    DtypeAny = np.dtype[tp.Any] # pylint: disable=W0611 #pragma: no cover
+    TNDArrayAny = np.ndarray[tp.Any, tp.Any] # pylint: disable=W0611 #pragma: no cover
+    TDtypeAny = np.dtype[tp.Any] # pylint: disable=W0611 #pragma: no cover
+    TSeriesAny = Series[tp.Any, tp.Any] # pylint: disable=W0611 #pragma: no cover
+    TFrameAny = Frame[tp.Any, tp.Any, tp.Unpack[tp.Tuple[tp.Any, ...]]] # type: ignore[type-arg] # pylint: disable=W0611 #pragma: no cover
```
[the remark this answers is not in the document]
flexatone: Maybe. The issue is that we still support old NumPy, pre generic `ndarray`, and this these have to be in a conditional if type-checking clause. You might be able to import from one if type checking to another if type checking, but it feels very strange.

static_frame/core/frame.py
```diff
@@ -3280,7 +3292,7 @@ def __init__(self,
         #-----------------------------------------------------------------------
         # blocks assignment
 
-        blocks_constructor: tp.Optional[tp.Callable[[ShapeType], None]] = None
+        blocks_constructor = _NA_BLOCKS_CONSTRCTOR
```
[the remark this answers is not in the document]
flexatone: pyright was not happy with using a `tp.Callable` here!

static_frame/core/frame.py
```diff
@@ -9203,9 +9227,12 @@ def to_latex(self,
                 )
         return repr(self.display(config))
 
+# NOTE: must not use decorator to avoid mypy confusion on interface
+doc_update(Frame.__init__, selector='container_init', class_name='Frame')
+
 #-------------------------------------------------------------------------------
 
-class FrameGO(Frame):
+class FrameGO(Frame[TVIndex, TVColumns, tp.Unpack[TVDtypes]]): # type: ignore[type-arg]
```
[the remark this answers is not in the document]
flexatone: Yes, that is a problem. It is also a problem with a list:

```python
x: list[str] = ['a', 'b']
x.append(None)
```
But our type checkers should raise an error when we try to append a type not known; I am not sure if I can do something similar yet with `FrameGO`.

static_frame/core/index_base.py
```diff
@@ -245,13 +245,13 @@ def _sample_and_key(self: I,
             count: int = 1,
             *,
             seed: tp.Optional[int] = None,
```
chaburkland, line 247: At first I thought you weren't using `__future__.annotations`, but I see you. As such, how come you haven't opted for `int | None`?

static_frame/core/index_datetime.py
```diff
@@ -55,21 +56,20 @@
...
 
-class IndexDatetime(Index):
+class IndexDatetime(Index[np.datetime64]):
     '''
     Derivation of Index to support Datetime operations. Derived classes must define _DTYPE.
     '''
 
     STATIC = True
-    _DTYPE: DtypeAny # define in derived class
+    _DTYPE: TDtypeAny # define in derived class
```
chaburkland, line 65: Wouldn't this be `_DTYPE: np.dtype[np.datetime64]`? All `IndexDatetime` subclasses need to have this dtype family

static_frame/core/index_hierarchy.py
```diff
@@ -237,31 +243,31 @@ class IndexHierarchy(IndexBase):
...
-    _index_types: tp.Optional['Series'] # Used to cache the property `index_types`
-    _pending_extensions: tp.Optional[tp.List[tp.Union[SingleLabelType, 'IndexHierarchy', PendingRow]]]
+    _index_types: tp.Optional[Series[tp.Any, np.object_]] # Used to cache the property `index_types`
+    _pending_extensions: tp.Optional[tp.List[tp.Union[TSingleLabel, 'IndexHierarchy', PendingRow]]]
 
     # _IMMUTABLE_CONSTRUCTOR is None from IndexBase
     # _MUTABLE_CONSTRUCTOR will be defined after IndexHierarhcyGO defined
 
-    _INDEX_CONSTRUCTOR: IndexConstructor = Index
+    _INDEX_CONSTRUCTOR: tp.Callable[..., 'IndexBase'] = Index
```
chaburkland, line 259: How come it's not `TIndexCtor`?

static_frame/core/index_hierarchy.py
```diff
@@ -1337,7 +1347,7 @@ def size(self: IH) -> int:
         return self._blocks.size
 
     @property
-    def nbytes(self: IH) -> int:
+    def nbytes(self) -> int:
```
chaburkland, line 1350: I guess that type-checkers always correctly infer the type of the first argument when named `self`? I would have expected it be "better" to do `def nbytes(self: tp.Self) -> int`, but perhaps that is not the case!

static_frame/core/index_hierarchy.py
```diff
@@ -2186,12 +2196,12 @@ def _ufunc_axis_skipna(self: IH,
     def _ufunc_shape_skipna(self, *,
             axis: int,
             skipna: bool,
-            ufunc: UFunc,
-            ufunc_skipna: UFunc,
+            ufunc: TUFunc,
```
chaburkland, line 2199: It must be late, because it seems funny to me that people unfamiliar with this might read this as `tuff-unk`

static_frame/core/interface.py
```diff
@@ -1123,29 +1174,35 @@ def name_obj_iter(cls,
...
         instance = cls.get_instance(target=target)
 
-        for name_attr in dir(target.__class__): # get metaclass
-            if name_attr == 'interface':
-                # getting interface off of the class will recurse
-                yield name_attr, None, ContainerBase.__class__.interface #type: ignore
+        # for name_attr in dir(target.__class__): # get metaclass
+        #     if name_attr == 'interface':
+        #         # getting interface off of the class will recurse
+        #         yield name_attr, None, ContainerBase.__class__.interface #type: ignore
```
chaburkland, line 1180: Consider removing?

static_frame/core/node_iter.py
```diff
@@ -31,11 +31,20 @@
...
-
-FrameOrSeries = tp.TypeVar('FrameOrSeries', 'Frame', 'Series', 'Bus', 'Quilt', 'Yarn')
+    TNDArrayAny = np.ndarray[tp.Any, tp.Any] # pylint: disable=W0611 #pragma: no cover
+    # TDtypeAny = np.dtype[tp.Any] # pylint: disable=W0611 #pragma: no cover
+    TSeriesAny = Series[tp.Any, tp.Any] # pylint: disable=W0611 #pragma: no cover
+    TFrameAny = Frame[tp.Any, tp.Any, tp.Unpack[tp.Tuple[tp.Any, ...]]] # type: ignore[type-arg] # pylint: disable=W0611 #pragma: no cover
+    TBusAny = Bus[tp.Any] # pylint: disable=W0611 #pragma: no cover
+    TYarnAny = Yarn[tp.Any] # pylint: disable=W0611 #pragma: no cover
+
+TFrameOrSeries = tp.TypeVar('TFrameOrSeries',
```
chaburkland, line 41: `TFrameOrSeries` is a bit of a misnomer. Perhaps `TContainerAny`?

static_frame/core/node_iter.py
```diff
@@ -118,15 +127,15 @@ def _apply_iter_items_parallel(self,
 
         # use side effect list population to create keys when iterating over values
         func_keys = []
-        arg_gen: PoolArgGen
+        # arg_gen: PoolArgGen
```
chaburkland, line 130: FYI this is the only reference to `PoolArgGen`, so if this change is permanent that type can be removed

static_frame/core/node_iter.py
```diff
@@ -189,7 +198,7 @@ def apply_iter_items(self,
 
     @doc_inject(selector='apply')
     def apply_iter(self,
-            func: AnyCallable
+            func: TCallableAny
```
chaburkland, line 201: I think you could constrain this a bit more. Perhaps:
```python
func: TCallableOneArg | TCallableTwoArgs
```

static_frame/core/util.py
```diff
@@ -415,61 +412,60 @@ def depth_level_from_specifier(
...
         tp.Iterable[TLabel],
         tp.Iterable[tp.Sequence[TLabel]], # only for IndexHierarchy
         # tp.Type['IndexAutoFactory'],
         ]
 
-IndexConstructor = tp.Optional[tp.Union[tp.Callable[..., 'IndexBase'], tp.Callable[..., 'Index'], tp.Type['Index']]]
-IndexConstructors = tp.Union[IndexConstructor,
-        tp.Sequence[IndexConstructor],
-        tp.Iterator[IndexConstructor],
+TIndexCtor = tp.Union[tp.Callable[..., 'IndexBase'], tp.Callable[..., 'Index'], tp.Type['Index']]
```
chaburkland, line 428: Why not the additional specifications for `IndexHierarchy`?

static_frame/core/util.py
```diff
@@ -3299,13 +3295,13 @@ class PositionsAllocator:
     '''Resource for re-using a single array of contiguous ascending integers for common applications in IndexBase.
     '''
     _size: int = 1024 # 1048576
-    _array: NDArrayAny = np.arange(_size, dtype=DTYPE_INT_DEFAULT)
+    _array: TNDArrayAny = np.arange(_size, dtype=DTYPE_INT_DEFAULT)
```
chaburkland, line 3298: How come not `np.ndarray[np.int_]`?

chaburkland: Same for all other functions on this class

README.rst
```diff
@@ -47,13 +47,27 @@
...
+Among the many Python DataFrame libraries, StaticFrame is an alternative that prioritizes correctness, maintainability, and reducing opportunities for error. Key features include:
 
-A wide variety of table formats are supported, including input from and output to CSV, TSV, JSON, MessagePack, Excel XLSX, SQLite, HDF5, NumPy, Pandas, Arrow, and Parquet; additionally, output to xarray, VisiData, HTML, RST, Markdown, and LaTeX is supported, as well as HTML representations in Jupyter notebooks. Full serialization is also available via custom NPZ and NPY encodings, the latter supporting memory mapping.
+* An immutable data-model, providing memory efficiency and excllent performance
+* Use Python type hints to statically type a DataFrame, including the index, columns, and columnar types
+* Use type hints and a family of validators to provide run-time type and data validation
+* A hierarchical and consistent interface (avoiding the inconsistencies of Pandas)
+* Full support of all NumPy dtypes and datetime64 units
+* Interchange with Pandas, Arrow, Parquet, CSV, TSV, JSON, MessagePack, Excel XLSX, SQLite, HDF5, and NumPy
```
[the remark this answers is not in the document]
flexatone: Translate between...

static_frame/core/container.py
```diff
@@ -209,19 +216,19 @@ class ContainerOperandSequence(ContainerBase):
 
     __slots__ = ()
 
-    interface: 'Frame' # property that returns a Frame
-    # values: NDArrayAny
+    interface: TFrameAny # property that returns a Frame
```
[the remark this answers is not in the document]
flexatone: Yes, mypy requires this as (I think) it does not recognize default parameters for generics.

static_frame/core/index_base.py
flexatone: That syntax is only available in Python 3.9. I am still supporting Python 3.8, and it gets confusing to use some at runtime and some at hint time, so I default to old-style for now.

static_frame/core/index_hierarchy.py
flexatone: Yeah,  that seems odd. That type should be as narrow as possible, as it does not need to support the general case. So the right thing to do is just remove the hint (as done in similar args elsewhere) and let it be the type of the argument.

static_frame/core/index_hierarchy.py
flexatone: hahaha, tough funk!

static_frame/core/series.py
```diff
@@ -131,26 +131,36 @@
...
+
+    FrameType = tp.TypeVar('FrameType', bound=TFrameAny) # pylint: disable=W0611 #pragma: no cover
 
 
 #-------------------------------------------------------------------------------
-class Series(ContainerOperand):
+TVDtype = tp.TypeVar('TVDtype', bound=np.generic, default=tp.Any)
+TVIndex = tp.TypeVar('TVIndex', bound=IndexBase, default=tp.Any)
+
+def _NA_VALUES_CONSTRCTOR(count: int) -> None: ...
```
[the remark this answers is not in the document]
flexatone: since I am using ctor now, `_NA_VALUES_CTOR` is even better!

static_frame/core/store.py
```diff
@@ -167,9 +169,9 @@ def get_field_names_and_dtypes(*,
 
             # add index names in front of column names
             if include_columns:
-                field_names.extend(columns_values)
+                field_names.extend(columns_values) # pyright: ignore
```
chaburkland, line 172: Do you know why this is ignored? `field_names: tp.Sequence[TLabel]` and `columns_values: tp.Sequence[TLabel]`

static_frame/core/yarn.py
```diff
@@ -64,23 +70,23 @@ class Yarn(ContainerBase, StoreClientMixin):
             '_deepcopy_from_bus',
             )
 
-    _series: Series
+    _series: TSeriesAny
```
chaburkland, line 73: Why not `TSeriesObject`?

static_frame/core/yarn.py
```diff
@@ -38,21 +38,27 @@
...
+    TNDArrayAny = np.ndarray[tp.Any, tp.Any] # pylint: disable=W0611 #pragma: no cover
+    TDtypeAny = np.dtype[tp.Any] # pylint: disable=W0611 #pragma: no cover
 
+TSeriesAny = Series[tp.Any, tp.Any]
+TFrameAny = Frame[tp.Any, tp.Any, tp.Unpack[tp.Tuple[tp.Any, ...]]] # type: ignore[type-arg]
+TBusAny = Bus[tp.Any]
 
-class Yarn(ContainerBase, StoreClientMixin):
+#-------------------------------------------------------------------------------
+TVIndex = tp.TypeVar('TVIndex', bound=IndexBase, default=tp.Any)
```
chaburkland, line 59: Shouldn't bound be `np.object_`? You can't have a yarn of anything other than `sf.Frame` or `sf.Series`

static_frame/core/yarn.py
```diff
@@ -300,7 +306,7 @@ def iter_element_items(self) -> IterNodeNoArg['Yarn']:
     # common attributes from the numpy array
 
     @property
-    def dtype(self) -> DtypeAny:
+    def dtype(self) -> TDtypeAny:
```
chaburkland, line 309: `TDtypeObject`?

static_frame/core/index_hierarchy.py
flexatone: Right, you never need to do `self: tp.Self`!

static_frame/core/interface.py
flexatone: Yes!

static_frame/core/node_iter.py
flexatone: Oh wow, you are correct. It expanded incrementally without me noticing the name was out of date!

static_frame/core/node_iter.py
flexatone: Good catch!

static_frame/core/store.py
flexatone: I only had the fortitude for very limited exploration of such questions: the overall typing maintenance was so high on-boarding pyright I erred on the side of ignoring!

static_frame/core/yarn.py
flexatone: Right! In my first pass accommodating generics, I erred on the side of over any-ing, again in large part because the number of changes needed was insane, so efficiency was key.

static_frame/core/frame.py
```diff
@@ -2180,27 +2189,27 @@ def from_delimited(cls,
         if skip_header < 0:
             raise ErrorInitFrame('skip_header must be greater than or equal to 0')
 
-        fp = path_filter(fp) # normalize Path to strings
+        fpf = path_filter(fp) # normalize Path to strings
```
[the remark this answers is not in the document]
flexatone: I agree, but the scope of these is so small that I think the short name and comment are fine for now.

static_frame/core/index_datetime.py
flexatone: Yes, good improvement!

static_frame/core/util.py
flexatone: Good catch!

static_frame/core/index_datetime.py
flexatone: Yes, good find!

static_frame/core/yarn.py
flexatone: No, I don't think so. This typevar is for the `Index` of the `Yarn`, which can be any Index. Separately, thanks to your feedback, we are now using `_series: TSeriesObject`, which is what I think you were thinking of here.

static_frame/core/yarn.py
flexatone: Correct!

static_frame/core/frame.py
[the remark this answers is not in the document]
flexatone: Maybe `FrameGO` should not be generic to columnar dtypes?

static_frame/core/node_iter.py
flexatone: This seems desirable, but leads to other issues, so I am going to defer this one for now.

static_frame/core/util.py
flexatone: I think it is already covered by `tp.Callable[..., 'IndexBase']`...

PR #850, opened 2023-09-22
Support loading parquet files saved with pyarrow<1.0

michaeljpeters:
pyarrow<1.0 saved parquet files in a format that cannot be loaded by pyarrow>=1.0 unless you add `use_legacy_dataset=True` to the pyarrow.parquet.read_table call. 

This changes adds a fallback to use_legacy_dataset=True if the initial read_table call fails.

Related pyarrow issues:
https://github.com/apache/arrow/issues/32660
https://github.com/apache/arrow/issues/24407

### DIFF
--- a/static_frame/core/frame.py
+++ b/static_frame/core/frame.py
@@ -27,7 +27,7 @@ from arraykit import resolve_dtype
 from arraykit import resolve_dtype_iter
 from arraykit import split_after_count
 from numpy.ma import MaskedArray
-from pyarrow.lib import ArrowInvalid
+from pyarrow.lib import ArrowInvalid  # pylint: disable=E0611
 
 from static_frame.core.archive_npy import NPYFrameConverter
 from static_frame.core.archive_npy import NPZFrameConverter
--- a/static_frame/core/frame.py
+++ b/static_frame/core/frame.py
@@ -3151,7 +3151,7 @@ class Frame(ContainerOperand):
         except ArrowInvalid:
             # support loading parquet files saved with pyarrow<1.0
             # https://github.com/apache/arrow/issues/32660
-            table = pq.read_table(fp,
+            table = pq.read_table(fp,  # pragma: nocover
                     columns=columns_select,
                     use_pandas_metadata=False,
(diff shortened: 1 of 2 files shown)


flexatone:
Thanks, @michaeljpeters , for this PR. Just a few small things needed:

### 1.
Mypy is complaining about this: 
```Run inv quality
static_frame/core/frame.py:30: error: Skipping analyzing "pyarrow.lib": module is installed, but missing library stubs or py.typed marker  [import]
static_frame/core/frame.py:30: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 135 source files)
Error: Process completed with exit code 1.
```

I think you just need to add `pyarrow*` to the exclusions in the mypy.ini
```
[mypy-IPython,openpyxl,pandas,pyarrow,pytest,setuptools,sphinx.*,sphinx_rtd_theme,scipy.*,tables,xarray,xlsxwriter.*,arraymap,arraykit,frame_fixtures]
ignore_missing_imports = True
```
### 2.
We might be missing coverage on the except scenario, as we do not store an old parquet file to test against (and do not generally store such files for testing, but generate them on the fly). In this case, assuming you have manually tested this, using `#pragma: no cover` on the except scenario is fine.

michaeljpeters:
Thanks for the quick response! I've manually tested using static-frame to load the legacy parquet files successfully. I'll add those typing updates shortly.

static_frame/core/frame.py
```diff
@@ -27,6 +27,7 @@
 from arraykit import resolve_dtype_iter
 from arraykit import split_after_count
 from numpy.ma import MaskedArray
+from pyarrow.lib import ArrowInvalid  # pylint: disable=E0611
```
michaeljpeters, line 30: Should this be moved to a dynamic import like L3135?

flexatone: Yes, good catch. `pyarrow` is not a required dependency, so we want to put the import within the function call.

flexatone (approved):
Thanks for this improvement.

PR #859, opened 2023-11-10
Update ftyping.md

fhenslee:
Reviewing ftyping.md and suggesting edits.

### DIFF
--- a/doc/source/articles/ftyping.md
+++ b/doc/source/articles/ftyping.md
@@ -60,7 +60,7 @@ def process(f: Frame[
                 ]: ...
 ```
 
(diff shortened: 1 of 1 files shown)


fhenslee:
@flexatone, I've included my edits and suggestions in this PR. This is new to me, so please let me know if there is a better format to provide this feedback.

flexatone:
Awesome, thanks @fhenslee !

fhenslee:
Take this for what it's worth, Grammarly's suggestion for improvement is below. 

Consider including more information on the potential limitations or challenges of using these new tools and features, as well as any potential trade-offs that users may need to consider. This could help readers make more informed decisions about whether or not to adopt these tools in their own work.

PR #861, opened 2023-11-15
Fix Docs Search

michaeljpeters:
Search was hanging indefinitely because of a sphinx_rtd_theme configuration problem.

Related issue: https://github.com/readthedocs/sphinx_rtd_theme/issues/1452

Example failing:
image


flexatone:
Many thanks for this improvement!

PR #867, opened 2023-11-27
Support callables in `TypeClinic`

flexatone:


flexatone:
This is difficult as we have to compare hint to thints.

PR #882, opened 2024-01-04
Add `BinaryIO` typing support to `Frame.to_npz`

michaeljpeters:
`Frame.to_npz` and `Frame.from_npz` support BinaryIO objects, however its type hint had implied that it only supported Paths.

This change adds typing support (and a test) for BinaryIO objects.

REPL Example of existing functionality:

```
>>> from tempfile import TemporaryFile
>>> f1 = sf.Frame.from_dict_records([{2: 4}, {2: 8}])
>>> with TemporaryFile() as tf:
...     f1.to_npz(tf)
...     f2 = sf.Frame.from_npz(tf)
>>> f1
<Frame>
<Index> 2       <int64>
<Index>
0       4
1       8
<int64> <int64>
>>> f2
<Frame>
<Index> 2       <int64>
<Index>
0       4
1       8
<int64> <int64>
```

### DIFF
--- a/static_frame/test/unit/test_frame.py
+++ b/static_frame/test/unit/test_frame.py
@@ -18,8 +18,8 @@ from hashlib import sha256
 from io import StringIO
 from itertools import chain
 from itertools import repeat
-from tempfile import TemporaryFile
 from tempfile import TemporaryDirectory
+from tempfile import TemporaryFile
 
 import frame_fixtures as ff
 import numpy as np
--- a/static_frame/test/unit/test_frame.py
+++ b/static_frame/test/unit/test_frame.py
@@ -18,10 +18,12 @@ from hashlib import sha256
 from io import StringIO
 from itertools import chain
 from itertools import repeat
+from tempfile import TemporaryFile
 from tempfile import TemporaryDirectory
 
 import frame_fixtures as ff
 import numpy as np
+import pytest
 import typing_extensions as tp
 
 import static_frame as sf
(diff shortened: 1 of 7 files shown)


static_frame/test/unit/test_frame.py
```diff
@@ -10438,32 +10440,33 @@ def test_frame_to_npz_a(self) -> None:
...
+    @pytest.mark.skip("failing due to index name mismatch")
     def test_frame_to_npz_d(self) -> None:
         f1 = ff.parse('s(10,100)|v(int,str,bool,bool,float,float)').rename(
                 'foo', index='bar', columns='baz')
 
         with temp_file('.npz') as fp:
             f1.to_npz(fp)
             f2 = Frame.from_npz(fp)
-            f1.equals(f2, compare_dtype=True, compare_class=True, compare_name=True)
+            self.assertTrue(f1.equals(f2, compare_dtype=True, compare_class=True, compare_name=True))
```
michaeljpeters, line 10469: Looks like these tests were missing an `assert`. This test started failing due to a mismatch in index name. Just marked as skipped since I think this is out of scope.

NOTE: it looks like there are other places in this file that are missing `assertTrue`, at least in the to_npy tests.

flexatone: Thanks for isolating this!

flexatone:
Thanks for this PR. I just stumbled on this same shortcoming the other day. Can you see if you can fix the test `test_bus_npz_b` that is failing?

michaeljpeters:
Looks like the test was failing since I was using `isinstance` with `tp.IO` instead of `io.IOBase`.

flexatone (approved):
Many thanks for these excellent improvements.

PR #883, opened 2024-01-06
Alias ContainerBase at root level

ForeverWintr:
I think it's useful to export `ContainerBase`, so that users can do high level checks to see if an object is a static_frame container (i.e., `if isinstance(thing, sf.ContainerBase)`). 

Any objections to doing so?


flexatone:
Thanks for this PR; no objections, moving this into the public namespace is fine.

flexatone (approved):
Thanks for this contribution!

PR #887, opened 2024-01-19
`Bus.max_persist > 1` issue

flexatone:


static_frame/core/bus.py
```diff
@@ -739,135 +739,142 @@ def rehierarch(self,
...
+                    for label in target_labels[target_loaded]: # update LRU position
+                        self._last_accessed[label] = self._last_accessed.pop(label, None)
+                    labels_to_read = target_labels[~target_loaded]
+                else: # no targets are loaded
+                    labels_to_read = target_labels
+
+            else: # loaded_needed > max_persist:
+                # Need to load more than max_persist, so limit to last max_persist-length components. All other Frame, if loaded, will be deleted
+                # assert max_persist < len(target_labels)
+                target_labels = target_labels[-max_persist:] # type: ignore # pylint: disable=E1130
```
chaburkland, line 818: I think this is the optimization, right? If max_persist is shorter than the number of items needed, there's no reason to load the initial frames, since they will immediately be dropped

chaburkland (approved):
These are great improvements!

static_frame/core/bus.py
flexatone: Yes, and an important one. Previously, if the selection range was greater than `max_persist`, we would have uselessly loaded and then dropped (potentially many) `Frame`!

PR #921, opened 2024-03-29
More flexible `Yarn` indicies

flexatone:


flexatone:
One of the issues here is that the stored `IndexHierarchy` is retained and used by the Yarn to lookup values in each `Bus`; i.e., it permits selecting the `Bus` a `Frame` might reside. If three `Bus` are provided and they all have the same outer label, this will not work. One option is to store a different `IndexHierarchy` that uses `iloc` positions at all levels; this will free up the "public" index from these restrictions.

Issue #919, opened 2024-03-26
`iter_label` interface should be implemented in `IndexBase`


flexatone:
```
    @property
    def iter_label(self) -> IterNodeDepthLevel[tp.Any]:
```

PR #923, opened 2024-04-08
Add iter_label interface to IndexBase

gtamba:
Addresses #919

I had a couple of questions since it's my first PR:

- Is it okay to import `IndexBase` in `typing/test_index.py` from the core module? 
- Does the test achieve the desired coverage? (mypy fails upon removal of the code addition to `IndexBase`)

### DIFF
--- a/static_frame/core/index_base.py
+++ b/static_frame/core/index_base.py
@@ -17,6 +17,7 @@ from static_frame.core.display_config import DisplayFormats
 from static_frame.core.doc_str import doc_inject
 from static_frame.core.exception import ErrorInitIndex
 from static_frame.core.node_dt import InterfaceDatetime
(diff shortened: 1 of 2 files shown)


flexatone:
@gtamba:  just one thing to fix: the quality checks are complaining about trailing white space. Usually best to handle this by configuring your editor to automatically trim trailing white space on save.

gtamba:
> @gtamba: just one thing to fix: the quality checks are complaining about trailing white space. Usually best to handle this by configuring your editor to automatically trim trailing white space on save.

Ah my bad, I only ran unit/typing tests before committing this. Just pushed the changes, and hopefully this should pass.

flexatone (approved):
Nice work, thanks for this contribution.

PR #924, opened 2024-04-10
Fully decoupled Yarn index

flexatone:


static_frame/core/yarn.py
```diff
@@ -669,25 +729,42 @@ def shapes(self) -> TSeriesObject:
         Returns:
             :obj:`tp.Series`
         '''
-        return Series.from_concat((b.shapes for b in self._values),
-                index=self._index)
+        array = np.empty(shape=len(self._index), dtype=DTYPE_OBJECT)
+
```
flexatone, line 733: pre fetch shapes per bus

PR #966, opened 2024-08-19
Mypy==1.11.1

flexatone:


flexatone:
https://github.com/python/mypy/issues/17691

Issue #937, opened 2024-06-03
`ErrorInitSeries: dimensionality of final values not supported` when calling `loc_max` on a series with an IndexHierarchy


ForeverWintr:
**Description**

I'm getting an error when calling `loc_max` on a frame with an index hierarchy.

**Example**

```python
f = sf.Frame.from_elements([1, 2], index=sf.IndexHierarchy.from_labels([[1, 2], [3, 4]]))
# <Frame>
# <Index>                  0       <int64>
# <IndexHierarchy>
# 1                2       1
# 3                4       2
# <int64>          <int64> <int64>

f.loc_max() # Raises ErrorInitSeries

```

**Platform**

Run the following function (static-frame >= 0.8.1) and provide the results to define your platform and environment:

```python
>>> import static_frame as sf
>>> sf.Platform.display()
<Series: platform>
<Index>
platform           Linux-6.8.0-76060800daily20240311-generic-x86_64-with-glibc2.35
sys.version        3.11.4 (main, May 22 2024, 09:24:56) [GCC 11.4.0]
static-frame       2.6.0
numpy              1.24.3
pandas             1.5.3
xlsxwriter         3.1.0
openpyxl           3.0.9
xarray             <ModuleNotFoundError>
tables             3.8.0
pyarrow            12.0.1
msgpack            None
msgpack_numpy      None
<<U13>             <object>
```

flexatone:
Corrected as part of the associated PR.

PR #977, opened 2024-09-23
Adds support for IH when using (i)loc_min(max)

chaburkland:
@flexatone What are your thoughts on this solution to #937 ?

It's marked as a draft now because I didn't want to spend time implementing all the necessary unit tests if you reject the approach outright

If you like it, then I'll get it properly PR ready

### DIFF
--- a/static_frame/core/frame.py
+++ b/static_frame/core/frame.py
@@ -7380,41 +7380,37 @@ class Frame(ContainerOperand, tp.Generic[TVIndex, TVColumns, tp.Unpack[TVDtypes]
 
     def _container_from_index_values(
             self,
-            post: np.ndarray,
+            values: np.ndarray,
             axis: int,
-            ) -> TSeriesAny | TFrameAny:
+            ) -> TSeriesAny:
+
+        if values.ndim == 1:
+            post = values
+        else:
+            post = np.fromiter(
+                map(tuple, values),
+                count=len(values),
+                dtype=object,
+            )
+
         if axis == 0:
-            if self.index._NDIM == 1:
-                return Series(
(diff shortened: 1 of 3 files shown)


chaburkland:
Refs #937

flexatone:
Thanks, @chaburkland , for this PR: this needed to be addressed! 

I see that you are returning a `Frame` when calling `loc_max` on a  `Frame` with `IndexHierarchy`, but returning a `Series` when calling `loc_max` on a `Frame` with a 1D index. Is this too surprising? We could always return a `Series`; in the case of an `IndexHierarchy`, the Series could be one of tuples for the IH labels, which makes using those values for selection in the same IH straight forward. 

One advantage of what you have done is that the IH name become the column labels: was that your motivation to going to a Frame?

chaburkland:
@flexatone I think you're right that getting back a Frame is a bit surprising. Let me fix it :upside_down_face:

static_frame/core/frame.py
```diff
@@ -7378,6 +7378,35 @@ def sample(self,
...
+
+        if values.ndim == 1:
+            post = values
+        else:
+            post = np.fromiter(
+                map(tuple, values),
+                count=len(values),
+                dtype=object,
+            )
+            post.flags.writeable = False
```
flexatone, line 7395: I think you can use `arraykit.array_to_tuple_array` for an optimized version of this move.

flexatone (changes requested):
Thanks for this PR! Just one opportunity here to use an `arraykit` function.

flexatone (approved):
Thanks!

Issue #986, opened 2024-10-23
Reindexing with `IndexHierarchy` objects is very slow


chaburkland:
**Example**

```python
import static_frame as sf
import random

index = sf.IndexHierarchy.from_product(range(10_000), range(100))

ilocs = list(range(len(index)))
random.shuffle(ilocs)

sub_index = index.iloc[ilocs[:len(index) // 2]]

s = sf.Series.from_element("a", index=sub_index)

s.reindex(index)               # 8.15 s ± 1.09 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
sub_index.intersection(index)  # 343 ms ± 22.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
index.intersection(sub_index)  # 353 ms ± 78.5 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
Compare to normal index
```python
index = sf.Index(range(10_000 * 100))
sub_index = index.iloc[ilocs[:len(index) // 2]]
s = sf.Series.from_element("a", index=sub_index)

s.reindex(index)               # 245 ms ± 7.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
sub_index.intersection(index)  # 146 ms ± 13.1 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
index.intersection(sub_index)  # 142 ms ± 6.86 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

**Platform**
```python
<Series: platform>
<Index>
platform           Linux-5.4.0-196-generic-x86_64-with-glibc2.31
sys.version        3.11.10 (main, Sep  7 2024, 18:35:42) [GCC 9.4.0]
static-frame       2.13.2
numpy              1.25.2
pandas             2.0.3
xlsxwriter         3.1.9
openpyxl           3.1.2
xarray             0.21.1
tables             3.9.1
pyarrow            10.0.1
msgpack            1.0.4
msgpack_numpy      None
<<U13>             <object>

```

PR #988, opened 2024-10-23
Updates ic to use performant IH operations

chaburkland:
Refs #986

### DIFF
--- a/static_frame/core/container_util.py
+++ b/static_frame/core/container_util.py
@@ -1521,6 +1521,7 @@ def index_many_to_one(
                 compare_dtype=True,
                 compare_name=True,
                 compare_class=True,
+                skipna=False,
                 ):
             # compare dtype as result should be resolved, even if values are the same
             if (many_to_one_type is ManyToOneType.UNION
@@ -1564,7 +1565,7 @@ def index_many_to_one(
                 index_dtypes_arrays = []
 
         if mtot_is_concat:
-            # store array for each depth; unpack aligned depths with zip
+            # store array for each depth to preserve types; unpack aligned depths with zip
             arrays = [[index.values_at_depth(d) for d in range(depth_first)]] #type: ignore
         else: # NOTE: we accept type consolidation for set operations for now
             arrays = [index.values]
--- a/static_frame/core/index.py
+++ b/static_frame/core/index.py
@@ -757,7 +757,7 @@ class Index(IndexBase, tp.Generic[TVDtype]):
 
         Equivalent to: self.iter_label().apply(other._loc_to_iloc)
         '''
-        if self.__len__() == 0 or other.__len__() == 0:
+        if self.__len__() == 0:
             return EMPTY_ARRAY
 
         # Equivalent to: ufunc_unique1d_indexer(self.values)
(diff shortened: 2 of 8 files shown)


static_frame/test/unit/test_index.py
```diff
@@ -1644,6 +1649,13 @@ def test_index_iloc_map_b(self) -> None:
         with self.assertRaises(KeyError):
             ih1.iter_label().apply(ih2._loc_to_iloc)
 
+    def test_index_iloc_map_c(self) -> None:
+        ih1 = Index(['a', np.nan])
+        ih2 = Index([np.nan, 'a'])
+
+        post = ih1._index_iloc_map(ih2)
+        self.assertEqual(post.tolist(), [1, 0])  # What should this be?
```
chaburkland, line 1657: @flexatone What do you think this behavior should be? I'm really unfamiliar with the rules regarding support for `Index` objects with `nan`s

flexatone: NumPy's `np.nan` as single instance of a NaN value that can be used to match object identity:
```
>>> np.nan is np.nan
True
>>> np.nan is (np.nan+1)
False
```
Of course, normally we would not expect NaNs to have the same object id. But by using the np.nan, we actually get a Python object that can be reliably used in a dictionary and SF index, but only from that very same np.nan instance:
```
>>> d = {np.nan: 0, None: 1}
>>> d[np.nan]
0
>>> d[float('nan')]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: nan
```
So indeed, in the example you provide, that np.nans can map to each other just as if they were matching numbers. Of course, this will not work with any NaN. 

But in general, SF's index objects treat NaNs just as NaNs are treated in a dictionary, which may still be surprising, but is consistent.
```
>>> d = {np.nan: 0, None: 1, float('nan'): 2, float('nan'): 3}
>>> d[np.nan]
0
>>> d[float('nan')]
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
KeyError: nan
>>> d[list(d.keys())[-1]]
3
```

static_frame/core/container_util.py
```diff
@@ -1521,6 +1521,7 @@ def index_many_to_one(
                 compare_dtype=True,
                 compare_name=True,
                 compare_class=True,
+                skipna=False,
```
chaburkland, line 1524: @flexatone Note this minor bug fix

static_frame/core/index.py
```diff
@@ -757,9 +757,12 @@ def _index_iloc_map(self: I, other: I) -> TNDArrayAny:
 
         Equivalent to: self.iter_label().apply(other._loc_to_iloc)
         '''
-        if self.__len__() == 0:
+        if self.__len__() == 0 or other.__len__() == 0:
             return EMPTY_ARRAY
 
+        if self.dtype == DTYPE_OBJECT or other.dtype == DTYPE_OBJECT:
+            return self.iter_label().apply(other._loc_to_iloc, dtype=DTYPE_INT_DEFAULT)  #type: ignore [no-any-return]
+
```
chaburkland, line 765: @flexatone Sorry for the comment spam. This was my solution to the NaN issue. If either Index being compared are object dtypes, I fall back on the default behavior. I only use the optimized numpy route if the dtypes are defined

flexatone (approved):
A very nice improvement!

Issue #990, opened 2024-10-24
`sf.Quilt.to_frame()` fails when underlying bus index is `sf.IndexDate`


chaburkland:
**Example**
```python
f1 = sf.Frame(np.arange(9).reshape(3,3), columns=tuple("abc"))
f2 = sf.Frame(np.arange(9, 18).reshape(3,3), index=[3,4,5], columns=tuple("abc"))
bus = sf.Bus.from_dict({date(2024, 1, 1): f1, date(2024, 1, 2): f2}, index_constructor=sf.IndexDate)
sf.Quilt(bus, retain_labels=True).to_frame()
"""
ErrorInitIndex: Cannot create an `Index` with a `datetime64` array (with `dtype` datetime64[D] and including ['2024-01-01']); use a subclass (e.g. `IndexDate`) directly or as a constructor argument.
"""
```

**Platform**
```python
<Series: platform>
<Index>
platform           Linux-5.4.0-196-generic-x86_64-with-glibc2.31
sys.version        3.11.10 (main, Sep  7 2024, 18:35:42) [GCC 9.4.0]
static-frame       2.13.2
numpy              1.25.2
pandas             2.0.3
xlsxwriter         3.1.9
openpyxl           3.1.2
xarray             0.21.1
tables             3.9.1
pyarrow            10.0.1
msgpack            1.0.4
msgpack_numpy      None
<<U13>             <object>

```

PR #992, opened 2024-10-25
Use index_auto_factory for quilt ctors

chaburkland:
refs #990


flexatone (approved):
Very nice work here, and thanks for the additional coverage!

PR #996, opened 2024-11-08
939/unset index drop

afares97:


static_frame/core/frame.py
```diff
@@ -6657,13 +6658,18 @@ def unset_index(self, *,
...
-                # No recache is needed as it's not possible for an index to be GO
-                yield from self._index._blocks._blocks # type: ignore
+            if not drop:
+                if self._index.ndim == 1:
+                    yield self._index.values
+                else:
+                    # No recache is needed as it's not possible for an index to be GO
+                    yield from self._index._blocks._blocks # type: ignore
             for b in self._blocks._blocks:
                 yield b
```
flexatone, line 6674: Not part of your changes but this is better as:
```suggestion
            yield from self._blocks._blocks
```

static_frame/core/frame.py
```diff
@@ -6645,6 +6645,7 @@ def set_index_hierarchy(self,
 
     def unset_index(self, *,
             names: tp.Sequence[TLabel] = (),
+            drop: bool = False,
             # index_column_first: tp.Optional[tp.Union[int, str]] = 0,
```
flexatone, line 6649: Not part of your changes but we can remove this now.

static_frame/core/frame.py
afares97: Yes much cleaner. Will commit suggestion.

static_frame/core/frame.py
```diff
@@ -6657,68 +6657,78 @@ def unset_index(self, *,
             consolidate_blocks:
             columns_constructors:
         '''
+        # disallows specifying names with 'drop=True'
+        if drop and names:
```
flexatone, line 6661: Subtle but by doing this `is` check we implicitly check that the drop argument is a Boolean; that way another truthy value (like `'keep'`!) cannot be used.

```suggestion
        if drop is True and names:
```

static_frame/core/frame.py
```diff
@@ -6657,68 +6657,78 @@ def unset_index(self, *,
...
         else:
             block_gen = blocks
 
-        if not names:
-            names = self._index.names
+        # if not names:
+        #     names = self._index.names
+        if drop:
+            # When dropping the index, keep the existing columns without adding index names
+            columns, own_columns = self._columns, True
```
flexatone, line 6684: We can only own self._columns if this is not a `FrameGO`, i.e., it is purely immutable. We can substitute, the `STATIC` Boolean attr:
```suggestion
            columns, own_columns = self._columns, self.STATIC
```
Please add a test that specifically creates a `FrameGO`, calls this method, then grows (mutating columns) and then ensures the `Frame` resulting from this call does not have its columns mutated.

flexatone (changes requested):
Just a few additional changes.

static_frame/core/frame.py
afares97: Roger that.

static_frame/core/frame.py
afares97: Ah nice.

flexatone (approved):

flexatone:
@afares97 :  I corrected the typing issue; we just needed to a forward declaration show that the value is `None | IndexBase`.
I forgot to mention: we should as add `drop` to `unset_columns`: are up for that?

afares97:
> @afares97 : I corrected the typing issue; we just needed to a forward declaration show that the value is `None | IndexBase`. I forgot to mention: we should as add `drop` to `unset_columns`: are up for that?

Definitely.

afares97:
Thank you for fixing the typing issue. I have added the drop parameter to unset_columns. Will soon add similar tests to ensure correct logic implementation.

static_frame/core/frame.py
```diff
@@ -6877,63 +6886,67 @@ def unset_columns(self, *,
...
+        if drop is True and names:
+            raise RuntimeError("The `names` parameter cannot be used with `drop=True` because the column labels will not be included in the resulting Frame.")
 
-        # columns blocks are oriented as "rows" here, and might have different types per row; when moved on to the frame, types will have to be consolidated "vertically", meaning there is little chance of consolidation. A maximal decomposition might give a chance, but each ultimate column would have to be re-evaluated, and that would be expense.
+        if not names and drop is False:
+            names = self._columns.names
 
-        blocks = TypeBlocks.from_blocks(
+        if drop is True:
+            index, own_index, blocks = self._index, self.STATIC, self._blocks
```
flexatone, line 6896: As Index on the rows are always immutable, here you can assume owning the index.
```suggestion
            index, own_index, blocks = self._index, True, self._blocks
```

flexatone (changes requested):
Looks good; just one thing to change and we should be ready.

flexatone:
@afares97 : note the quality check failure:
```
static_frame/core/frame.py:6944: error: Incompatible types in assignment (expression has type "IndexBase | None", variable has type "IndexBase")  [assignment]
```
Is the same as before: we need to add a pre-usage variable type annotation like this, but for `index` now:
https://github.com/static-frame/static-frame/pull/996/commits/3140904fbb0e84fc77d6bbb1e2f6e36c69f6ae0c

afares97:
Changes are in and tests ran successfully. Thanks for the help.

flexatone:
@afares97 : a few quality things to check on:
https://github.com/static-frame/static-frame/actions/runs/11920048581/job/33221674742?pr=996

afares97:
> @afares97 : a few quality things to check on: https://github.com/static-frame/static-frame/actions/runs/11920048581/job/33221674742?pr=996

Done. I hope.

flexatone (approved):

flexatone:
All good! Thanks for the contribution, @afares97

PR #999, opened 2024-11-22
Implement `abs()` for all containers

yanbai123:


flexatone:
Thanks for these contributions. Here is what we need to do to fix the interface test and update the example generation routines.

1. In interface.py, update `UFUNC_UNARY_OPERATORS` to include 'abs'
2. If you now run `pytest -k test_interface_summary_b`, you should see the deviation now in 'Operator Unary', not 'Method'; you can now fix this test to show the right number of unary operators.
3. In build_example.py, for every class that has an 'operator_unary()' method, add an `elif` where `attr == 'abs'` and include an `s.abs()` statement.
4. To test this, run `python3 doc/test_example_gen.py`

Please let me know if you have questions.

doc/build_example.py
```diff
@@ -6387,6 +6407,9 @@ def operator_unary(row: sf.Series) -> tp.Iterator[str]:
         if attr == '__abs__()':
             yield f'bt = {icls}({kwa(BATCH_INIT_B)})'
             yield f'abs(bt).to_frame()'
+        elif attr == 'abs()':
+            yield f'bt = {icls}({kwa(BATCH_INIT_B)})'
+            yield f'bt.abs()'
```
flexatone, line 6412: To show the result of using the `abs()` here, we need to call `to_frame()`, so:
```suggestion
            yield f'bt.abs().to_frame()'
```

flexatone (changes requested):
Overall, this looks great.

Lets add one more test of `Frame.abs()`: it will be there, but good to exercise other classes.

flexatone (approved):
Thanks, @yanbai123 , for this contribution!

PR #1074, opened 2025-06-16
Burkland/doc build task bug

chaburkland:
The task `doc`, attempts to execute a file that doesn't exist in the repo. I've updated the task, and added a symlink to the old file. Not sure if that's needed


flexatone (approved):
Great, thanks!

PR #1075, opened 2025-06-16
Updates a signature typehint. Removes a lot of unnecessary ignores

chaburkland:
Closes https://github.com/static-frame/static-frame/issues/961


flexatone (approved):
Nice improvement!

PR #1076, opened 2025-06-18
Fixes bug & adds test

### LINKED ISSUE #1073: `IndexHierarchy` set operation failure for object mappings
**Description**
When an `IndexHierarchy` object is sufficiently large enough (either through depth, size, or some combination thereof), the internal encoding must be represented as a Python integer, to handle encodings beyond 64 bits.

There is an incorrect assertion in the code that a code path requires the encoding to be in uint64, when that is not actually the case

**Example**

```python
# There might be a way to more easily create this, but I found this to be fast and small enough to demonstrate the issue
ih1 = sf.IndexHierarchy.from_labels(np.arange(0,     1_300).reshape(130, 10))
ih2 = sf.IndexHierarchy.from_labels(np.arange(1_300, 2_600).reshape(130, 10))
ih1.intersection(ih2)  # AssertionError!
ih1.union(ih2)  # AssertionError!
ih1.difference(ih2)  # AssertionError!
```

**Platform**
```python
<Series: platform>
<Index>
platform           Linux-5.4.0-216-generic-x86_64-with-glibc2.31
sys.version        3.11.4 (main, Aug 14 2024, 14:18:46) [GCC 9.4.0]
static-frame       3.1.0
numpy              1.24.3
pandas             1.5.3
xlsxwriter         3.2.2
openpyxl           3.1.5
xarray             <ModuleNotFoundError>
pyarrow            14.0.2
<<U12>             <object>

```

chaburkland:
Closes #1073

### DIFF
--- a/static_frame/core/loc_map.py
+++ b/static_frame/core/loc_map.py
@@ -462,12 +462,11 @@ class HierarchicalLocMap:
     @staticmethod
     def encode(indexers: TNDArrayAny, bit_offset_encoders: TNDArrayAny) -> TNDArrayAny:
         """
-        Encode indexers into a 1-dim array of uint64
+        Encode indexers into a 1-dim array
         """
         # Validate input requirements
         assert indexers.ndim == 2
         assert indexers.shape[1] == len(bit_offset_encoders)
-        assert indexers.dtype == DTYPE_UINT_DEFAULT
 
         array: TNDArrayAny = np.bitwise_or.reduce(indexers << bit_offset_encoders, axis=1)
         return array
@@ -549,7 +548,11 @@ class HierarchicalLocMap:
         starts = bit_offset_encoders
         stops: TNDArrayAny = np.empty(len(starts), dtype=dtype)
         stops[:-1] = starts[1:]
-        stops[-1] = 64
+
+        if bit_offset_encoders[-1] < 64:
+            stops[-1] = 64
+        else:
+            stops[-1] = encoded_arr.max().bit_length()
 
         lens = stops - starts
         masks = [x for x in (1 << lens) - 1]
--- a/static_frame/test/property/test_util.py
+++ b/static_frame/test/property/test_util.py
@@ -308,11 +308,11 @@ class TestUnit(TestCase):
         self.assertTrue(post.ndim == 1)
 
         # Coerce NaTs to Nones to ensure set operations work as expected
-        if resolve_dtype(*(arr.dtype for arr in arrays)).kind == 'O':
-            for i in range(len(arrays)):
-                arr = arrays[i]
-                if arr.dtype.kind in 'Mm' and np.isnat(arr).any():
-                    arrays[i] = arrays[i].astype(object)
+        try:
+            if any(np.isnat(arr).any() for arr in arrays):
+                return
+        except TypeError:
+            pass
 
         # nan values in complex numbers make direct comparison tricky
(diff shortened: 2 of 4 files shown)


static_frame/test/property/test_util.py
```diff
@@ -306,6 +306,14 @@ def test_intersect1d(self, arrays: tp.Sequence[np.ndarray]) -> None:
...
         post = util.intersect1d(arrays[0], arrays[1], assume_unique=False)
         self.assertTrue(post.ndim == 1)
+
+        # Coerce NaTs to Nones to ensure set operations work as expected
+        if resolve_dtype(*(arr.dtype for arr in arrays)).kind == 'O':
+            for i in range(len(arrays)):
+                arr = arrays[i]
+                if arr.dtype.kind in 'Mm' and np.isnat(arr).any():
+                    arrays[i] = arrays[i].astype(object)
+
```
chaburkland, line 316: This was caught by hypothesis, but it is unrelated to my PR, as it also fails on default (as of d561ee05d686269507ca160a314c35b908d11232)

The falsifying example was `@reproduce_failure('6.92.1', b'AAAFAAIBAAABAIAAAAAAAAAAAgEA')`

`arrays` is `[array(['NaT'], dtype='datetime64[D]'), array([None], dtype=object)]`

static_frame/test/property/test_util.py
```diff
@@ -306,6 +306,14 @@ def test_intersect1d(self, arrays: tp.Sequence[np.ndarray]) -> None:
...
         post = util.intersect1d(arrays[0], arrays[1], assume_unique=False)
         self.assertTrue(post.ndim == 1)
+
+        # Coerce NaTs to Nones to ensure set operations work as expected
+        try:
+            if any(np.isnat(arr).any() for arr in arrays):
+                return
+        except TypeError:
+            pass
+
```
chaburkland, line 316: See #1079

flexatone (changes requested):

static_frame/core/loc_map.py
```diff
@@ -549,7 +548,11 @@ def unpack_encoding(
         starts = bit_offset_encoders
         stops: TNDArrayAny = np.empty(len(starts), dtype=dtype)
         stops[:-1] = starts[1:]
-        stops[-1] = 64
+
+        if bit_offset_encoders[-1] < 64:
+            stops[-1] = 64
+        else:
+            stops[-1] = encoded_arr.max().bit_length()
```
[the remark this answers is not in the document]
chaburkland: This arises when encoding overflows!
Say we have 10 depths, and each depth's max value requires 8 bits.
This means that the encoded value is 80 bits!

This portion of the algorithm is unpacking an encoded number into its consituent parts. So if we see that the the largest number in the arr is 80 bits, and we see the offset encoders are:
`[0,10,20,30,40,50,60,70]`, when calculating the "stops", we shift left, and then assign the max bit length of the final value to largest possible size, resulting in `[10,20,30,40,50,60,70,80]`.

I previously hardcoded it to 64, which is always correct if the encoded number is not a Python int.

static_frame/test/property/test_util.py
chaburkland: Hypothesis found a new issue with this test, that is completely unrelated to my change. This also fails on default, and since I didn't want to fix it in this PR, I skipped it for now and created a ticket. The comment is misledading - I'll fix it now

flexatone (approved):
Thanks for this improvement!

PR #1078, opened 2025-06-21
391/index is sorted

chaburkland:


static_frame/core/util.py
```diff
@@ -660,6 +660,35 @@ class IterNodeType(Enum):
     ITEMS = 2
 
 
+class SortStatus(Enum):
+    ASC = 1
+    DESC = 2
+    NO = 3
```
flexatone, line 666: Small quibble: `NO` feels like a different tense than `ASC`  and `DESC`; maybe "NA" or "UNSORTED" or something else...

chaburkland: I wondered about this haha. I like `NA` best!

static_frame/core/bus.py
```diff
@@ -1582,14 +1585,31 @@ def sort_index(
...
+            kind=kind,
+            check=False,  # do not check at this level
+            no_ordering=True,  # we can't use ordering - don't calculate it!
+        )
+
+        if prep.behavior is prep.Behavior.RETURN_INDEX:
+            return self.__copy__()
+
+        if prep.behavior is prep.Behavior.REVERSE_INDEX:
+            return self._extract_iloc(REVERSE_SLICE)
```
chaburkland, line 1606: If we are sorted in the opposite direction, then we can just reverse the index!

.coveragerc
```diff
@@ -0,0 +1,4 @@
+[report]
+
+exclude_also =
+    if tp.TYPE_CHECKING:
```
chaburkland, line 4: I noticed you added `# pragma: no cover` on every line inside type-checking blocks. This config file specification means you don't have to do that!

static_frame/core/container_util.py
```diff
@@ -1822,14 +1815,13 @@ def prepare_values_for_lex(
     """Prepare values for lexical sorting; assumes values have already been collected in reverse order. If ascending is an element and values_for_lex is None, this function is pass through."""
     asc_is_element = isinstance(ascending, BOOL_TYPES)
     if not asc_is_element:
-        ascending = tuple(ascending)  # type: ignore
```
chaburkland, line 1825: This was needed before because you had support for `tp.Iterable[bool]`, but per our discussion, the signature now requires sized containers of bool

.coveragerc
flexatone: Awesome! I wish I had this earlier...

static_frame/core/util.py
```diff
@@ -622,6 +649,60 @@ class IterNodeType(Enum):
...
+        - `SortStatus.DESC` iif all flags are False
+        - `SortStatus.UNKNOWN` else
+
+        If key is not None, return `SortStatus.UNKNOWN`
+        If kind is not 'mergsort', return `SortStatus.UNKNOWN`
+        """
+        if key is not None or kind != DEFAULT_SORT_KIND:
+            return cls.UNKNOWN
+
+        if not isinstance(ascending, bool):
```
flexatone, line 698: I am not sure you should do this, but I had to test it:
```
>>> timeit isinstance(True, bool)
9.49 ns ± 0.148 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)
>>> timeit True is True or True is False
7.46 ns ± 0.5 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)
```

static_frame/core/util.py
```diff
@@ -3925,3 +4006,38 @@ def iloc_to_insertion_iloc(
...
+            # NOTE: cannot get view of object; use string
+            consolidated = view_2d_as_1d(group.astype(str))
+        else:
+            consolidated = view_2d_as_1d(group)
+        transitions = nonzero_1d(consolidated != roll_1d(consolidated, 1))[1:]
+    else:
+        group_to_tuple = False
+        transitions = nonzero_1d(group != roll_1d(group, 1))[1:]
+
+    return _slices_from_transitions(transitions, len(group)), group_to_tuple  # type: ignore
```
flexatone, line 4043: This could be implemented in ArrayKit at maybe substantial savings: integers from transitions could be read directly in C and used to create and yield slices, with all integer management in C. maybe create an issue in ArrayKit for this.

static_frame/core/sort_interface_mixin.py
```diff
@@ -0,0 +1,35 @@
...
+    def _apply_ordering(
+        self,
+        order: TNDArrayIntDefault,
+        sort_status: SortStatus,
+        axis: int = 0,
+    ) -> tp.Self:
+        """
+        Return a copy of this container with the specified ordering applied along the index of axis
+        """
+        raise NotImplementedError()  # pragma: no cover
```
flexatone, line 32: I think this should be a Protocol instead of a concrete class. As is, this will live in the MRO and incur some run-time cost. If this is a Protocol, I believe it will be removed entirely at runtime and incur zero cost.

flexatone (changes requested):
Thanks for this excellent contribution to performance, with some extensive (and noisy!) typing cleanup. 

I have a few suggestions for more reuse, but probably the thing that is most relevant is to implement `SortInterfaceMixin` as a `Protocol` rather than a class. As it has no implementations, this will provide a true zero-cost typing interface without complicating the MROs of all major containers.

static_frame/core/index_base.py
```diff
@@ -223,6 +229,39 @@ def index_types(self) -> Series[tp.Any, np.object_]:
...
+        """
+        sort_status = SortStatus.from_sort_kwargs(ascending, key, kind)
+        reportable_sort = sort_status is not SortStatus.UNKNOWN
+
+        if reportable_sort and self._sort_status is not SortStatus.UNKNOWN:
+            return True
+
+        order = sort_index_for_order(self, kind=kind, ascending=ascending, key=key)
+
+        is_sorted = order_is_sorted_asc(order)
```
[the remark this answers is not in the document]
chaburkland: I'll add a comment as this is non-intuitive!

Basically, when you call `sort_index_for_order`, the values you get back are always going to be oriented in the `ascending=True` direction, _from the perspective of the sort params_

If the index is already sorted according to the params, then the ordering array you get back will simply be 0->n. If you get back anything else, it means the arrays is NOT sorted according to the params.

That's why this only needs to check a single direction.

static_frame/core/index_datetime.py
```diff
@@ -267,7 +278,11 @@ def from_date_range(
             dtype=DT64_YEAR,
         )
         labels.flags.writeable = False
-        return cls(labels, name=name)
+        return cls(
+            labels,
+            name=name,
+            sort_status=SortStatus.from_bool(step >= 1),
```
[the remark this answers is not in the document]
chaburkland: I'll add this!

static_frame/core/index_hierarchy.py
```diff
@@ -1500,12 +1515,14 @@ def _drop_iloc(
...
         blocks = TypeBlocks.from_blocks(self._blocks._drop_blocks(row_key=key))
 
-        return self.__class__._from_type_blocks(
+        index = self.__class__._from_type_blocks(
             blocks=blocks,
             index_constructors=self._index_constructors,
             name=self._name,
             own_blocks=True,
         )
+        index._sort_status = self._sort_status
```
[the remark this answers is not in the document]
chaburkland: Great call

static_frame/core/util.py
chaburkland: Haha I like it

static_frame/core/container_util.py
```diff
@@ -1919,3 +1911,59 @@ def iter_component_signature_bytes(
...
+@tp.overload
+def sort_index_from_params(
+    index: IndexBase,
+    ascending: TBoolOrBools,
+    key: tp.Any,
+    kind: TSortKinds,
+    container: TSortInterface,
+    *,
+    axis: int = 0,
+    apply_ordering: tp.Literal[True] = True,
```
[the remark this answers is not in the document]
chaburkland: @flexatone Pyright gets mad when I remove the default:

https://github.com/static-frame/static-frame/actions/runs/17112700491/job/48537406122?pr=1078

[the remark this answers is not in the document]
flexatone: interesting; no problem then to leave it!

static_frame/core/sort_interface_mixin.py
chaburkland: Per our discussion, I will remove the subclass and opt for just duck-typing with type-ignores. I'll add an issue to explore metaclass resolution in a different PR! This will hopefully resolve the long standing issue of allowing us to add `abc.Sized` in the class MRO!

flexatone (approved):

PR #1084, opened 2025-06-30
Free-threaded performance study

flexatone:


brandtbucher:
Sorry to barge in, but I saw this come through my feed (I still subscribe to this repo). :)

With my "CPython maintainer" hat on, I'd be curious to hear about your experience porting something like StaticFrame to support free-threading, what issues (if any) you've run into, what performance impact or scaling bottlenecks you're seeing, etc. A large-ish production library like SF with non-trivial third-party dependencies and C extensions makes an excellent case-study!

brandtbucher:
(I realize the work may still be early/in-progress, too.)

flexatone:
Hi @brandtbucher , very nice to hear from you. When I learned at PyCon that NumPy had free-threaded compatible wheels available, I was inspired to get ArrayKit compatible with at least building 3.13t wheels. That was not too hard; I may still have thread safety issues, but I was able to get those wheels published. 

Given that SF (a) is built on immutable arrays and (b) has had primitives for doing multi-threaded function application for a long time, I was eager to see if there was a short path to performance. Amazingly, I found it!

Here is my first panel study doing row-wise function application on `Series`;  the function is trivial and still mostly NumPy-bound `lambda s: ((s % 2) == 0).sum()`. Even so, I am getting 2x performance with threads. In most real-world cases, the per-row function is heavier on Python, and I would expect those cases to perform even better.

This is the easiest performance win of this magnitude I have ever encountered! 

ft-perf

flexatone:
The out-performance is even better on MacOS...
ft-perf

brandtbucher:
Neat!

It's interesting how "wide" scales better than "tall". It makes sense that when the units of work are smaller, and there's more of them, that accessing the frame itself becomes the bottleneck and more threads just make things worse.

On the other hand, while "wide" still doesn't scale well as more threads are added, it at least doesn't seem to suffer the way "tall" does. The exception is "uniform wide" of course, which looks like it scales very well. I wonder why?

brandtbucher:
Cool, thanks for the link! I'd also appreciate it if you're able to ping me on any thread-safety issues/PRs you encounter going forward, too.

PR #1091, opened 2025-08-25
fix syntax error in `ftyping.py` dtype declaration

jorenham:


flexatone (approved):
Many thanks for finding this!

PR #1094, opened 2025-09-09
1088/abc support

chaburkland:


pyproject.toml
```diff
@@ -66,6 +66,7 @@ ignore = ["F841",
     "S310", # url schemes
     ]
 select = [
+    "I",   # isort
```
chaburkland, line 69: Cleans up how you are doing `ruff` in `tasks.py`!

static_frame/core/series_mapping.py
```diff
@@ -39,13 +39,13 @@ def __contains__(
         # linear time unavoidable
         return self._values.__contains__(key)
 
-    def __iter__(self) -> Iterator[TVValues]:
+    def __iter__(self) -> tp.Iterator[TVValues]:
```
chaburkland, line 42: Note the change from `abc.Iterator` -> `tp.Iterator`

static_frame/test/unit/test_index_base.py
```diff
@@ -7,42 +7,7 @@
...
-
-        with self.assertRaises(NotImplementedError):
-            idx1.copy()
-
-        with self.assertRaises(NotImplementedError):
-            idx1.display()
-
-        with self.assertRaises(NotImplementedError):
-            idx1.from_labels(())
+    pass
```
chaburkland, line 10: This is no longer possible given that `IndexBase` is now an incomplete ABC class, and so you can't even instantiate `IndexBase()`

tasks.py
```diff
@@ -157,19 +157,18 @@ def pyright(
     context.run('pyright', pty=pty)
 
 
-@task
-def format_check(context):
-    """Run formatting checks."""
-    context.run('ruff check --select I')
```
chaburkland, line 163: This is unnecessary! Instead, by adding "I" to the toml file, you now get this normally just using `ruff format --check`

PR #1105, opened 2025-10-16
`Bus.relabel` optimization & misc cleanup

### LINKED ISSUE #1104: `sf.Bus.relabel` on a store-backed bus requires eagerly loading everything
**Description**

The implementation of `sf.Bus.relabel` is not efficient if the bus is lazily backed on a store.

A mental model of `relabel` is that it should be an extremely-fast operation, where the only time-complexity is related to the length of the Index and any complexity provided by a users callable.

In the case of a lazy bus though, the implementation forces a read of every frame from the store, which is heavy and unnecessary for the operation

**Platform**

```python
<Series: platform>
<Index>
platform           Linux-6.14.0-33-generic-x86_64-with-glibc2.39
sys.version        3.11.4 (main, May  7 2025, 11:43:34) [GCC 13.3.0]
static-frame       3.3.0
numpy              2.2.4
pandas             2.2.3
xlsxwriter         3.2.2
openpyxl           3.1.5
xarray             <ModuleNotFoundError>
pyarrow            19.0.1
<<U12>             <object>

```

chaburkland:
Closes #1104 

1. Move common relabeling code to shared util func
2. Optimize path in `sf.Bus.relabel` when the bus is a partially loaded store-backed bus

### DIFF
--- a/static_frame/core/bus.py
+++ b/static_frame/core/bus.py
@@ -1072,7 +1072,7 @@ class Bus(ContainerBase, StoreClientMixin, tp.Generic[TVIndex]):
             else:
                 labels_to_load = index.values[labels_unloaded]
 
-            to_load_count = len(labels_to_load)  # type: ignore
+            to_load_count = len(labels_to_load)  # pyright: ignore
             if to_load_count + loaded_count <= max_persist:
                 for label, idx, f in zip(
                     labels_to_load,
--- a/static_frame/core/node_iter.py
+++ b/static_frame/core/node_iter.py
@@ -55,7 +55,7 @@ TContainerAny = tp.TypeVar(
         ' | Bus[tp.Any]'
         ' | Quilt'
         ' | Yarn[tp.Any]'
-    ), # pyright: ignore[reportGeneralTypeIssues]
+    ),  # pyright: ignore[reportGeneralTypeIssues]
 )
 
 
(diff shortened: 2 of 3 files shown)


static_frame/core/bus.py
```diff
@@ -768,9 +769,37 @@ def relabel(
...
+            # Match the persistent status
+            relabeled._values_mutable = self._values_mutable.copy()
+            relabeled._loaded = self._loaded.copy()
+            relabeled._loaded_all = self._loaded_all
+
+            if self._max_persist is not None:
+                relabeled._last_loaded = {
+                    relabeled.index[self._index.loc_to_iloc(self_label)]: None
+                    for self_label in self._last_loaded.keys()
+                }
```
flexatone: This is a little sketchy, not because it is incorrect, but because we are manipulating a bundle of private state far from the implementation. I wonder if it is better to implement this as a private method on the `Bus`... also, are there other places where we might do this kind of move that we have not been? If there are, having a dedicated method makes even more sense.

This will also need a bit of testing to make sure we have covered all the relevant cases; as you know, `Bus` persistence is tricky.

Also, should this be `indexi_init` used here and not `self._index`?
```
                relabeled._last_loaded = {
                    relabeled.index[index_init.loc_to_iloc(self_label)]: None
                    for self_label in self._last_loaded.keys()
                }
```

PR #1159, opened 2026-05-04
Mp context in `get_concurrent_executor`

ForeverWintr:
In [util.py](https://github.com/static-frame/static-frame/blob/4d64f47d822bcd4cb162ba699e7c86aa6638404e/static_frame/core/util.py#L1062), the `mp_context` argument for `get_concurrent_executor` is typehinted as a string, but is passed to `ProcessPoolExecutor`, which expects a `multiprocessing.context.BaseContext` instance:

```python
TMpContext: tp.TypeAlias = tp.Literal['fork', 'forkserver', 'spawn'] | None


def get_concurrent_executor(
    *,
    use_threads: bool,
    max_workers: tp.Optional[int],
    mp_context: TMpContext,
) -> tp.Type[Executor]:
    # NOTE: these imports are conditional as these modules are not supported in pyodide
    exe: tp.Callable[..., Executor]
    if use_threads:
        from concurrent.futures import ThreadPoolExecutor

        exe = partial(ThreadPoolExecutor, max_workers=max_workers)
    else:
        from concurrent.futures import ProcessPoolExecutor

        exe = partial(ProcessPoolExecutor, max_workers=max_workers, mp_context=mp_context)  # pyright: ignore
    return exe  # type: ignore
```

This PR updates the function to call `multiprocessing.get_context(mp_context)`.

### DIFF
--- a/static_frame/core/util.py
+++ b/static_frame/core/util.py
@@ -20,7 +20,6 @@ from io import StringIO
 from itertools import chain, zip_longest
 from os import PathLike
 
-
 import numpy as np
 import typing_extensions as tp
 from arraykit import (
--- a/static_frame/core/util.py
+++ b/static_frame/core/util.py
@@ -20,6 +20,7 @@ from io import StringIO
 from itertools import chain, zip_longest
 from os import PathLike
 
+
 import numpy as np
 import typing_extensions as tp
 from arraykit import (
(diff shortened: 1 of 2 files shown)


flexatone (approved):
Many thanks for this improvement!

